PR #34, opened 2022-02-23
5.0.0 proposal

zstewar1:
Removes the `SerialPort` trait, and instead creates a `struct SerialPort` which wraps and hides platform-specific serial port implementions (`crate::sys::SerialPort`). This removes the distinction between serial port trait objects and platform specific serial port structs, and prevents the need for using `Box<dyn SerialPort>` to write platform-agnostic code, since the `SerialPort` struct just compiles to the platform specific type. Platform specific behaviors are exposed as extension traits which only appear on the appropriate platform.

This PR is mostly ready, but may still require some documentation updates. It also has not be tested thoroughly at this point, since I don't have access to a mac and haven't had time to do tests on windows or linux. If this proposal seems good, I can go ahead and test it a bit more thoroughly.


jessebraham:
Thanks for the PR, sorry for the late response.

Just to set your expectations, I likely won't be getting to this for awhile. I'm still working on getting all the details of the fork settled and would also like to release at least one more `4.x.x` version first. I promise to come back to this at some point though!

PR #36, opened 2022-03-01
Fix filehandle leak in open() on Windows

sionik:
This fixes a filehandle leak on Windows when the call to `set_dcb()` does not work properly.


jessebraham (approved):
Sneaky! Thanks for fixing this

PR #39, opened 2022-03-10
Make sure fd is properly closed if initialization fails

### LINKED ISSUE #38: Possible file descriptor leak in open
The early return in https://github.com/serialport/serialport-rs/blob/ea7b56329c40e6deb8da7691a662b6d5c8974dcc/src/posix/tty.rs#L117 will exit the whole function, not just the block https://github.com/serialport/serialport-rs/blob/ea7b56329c40e6deb8da7691a662b6d5c8974dcc/src/posix/tty.rs#L95-L129.
Therefore, the call to `close` in https://github.com/serialport/serialport-rs/blob/ea7b56329c40e6deb8da7691a662b6d5c8974dcc/src/posix/tty.rs#L131 won't be executed.

jannic:
Closes #38

### DIFF
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -144,7 +144,7 @@ impl TTYPort {
         #[cfg(not(any(target_os = "ios", target_os = "macos")))]
         termios::set_baud_rate(&mut termios, builder.baud_rate);
         #[cfg(any(target_os = "ios", target_os = "macos"))]
-        termios::set_termios(fd, &termios, builder.baud_rate)?;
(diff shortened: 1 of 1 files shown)


jannic:
> It seems there is one instance of `termios::set_termios` being invoked which did not get updated, once that's fixed I'm happy to merge.

Oops, yes, sorry. Missed that one as it's cfg-gated on ios/macos.

jessebraham (approved):
Not a problem, that's why we have CI :grin: Thanks for getting that fixed!

Issue #37, opened 2022-03-08
tcflush on open may cause data loss


jessebraham:
This issue was migrated from GitLab. The original issue can be found here:
https://gitlab.com/susurrus/serialport-rs/-/issues/117
---
Currently, serialport-rs issues a `unsafe { tcflush(fd, libc::TCIOFLUSH) };` shortly after opening the serial device.
https://gitlab.com/susurrus/serialport-rs/-/blob/master/src/posix/tty.rs#L123

Opening the device already asserts DTR, so there is a small window where DTR is high but data received in that window will get lost.

This was noticed with an embedded device which starts sending serial data immediately after detecting DTR. It was reproduced on both Linux and MacOS. It probably affects all systems using `posix/tty.rs`.

In this case, the firmware of the embedded device is open source, so inserting a small delay between detecting DTR and sending data worked around the issue. But in general, I think if a host asserts DTR, it should really be ready to receive data.

Unfortunately, it seems to be impossible to flush the buffers before opening the device, and it's also impossible to open the device without asserting DTR. (See for example https://unix.stackexchange.com/questions/446088/how-to-prevent-dtr-on-open-for-cdc-acm)

Therefore, the only solution would be to remove the call to `tcflush` (or make it optional, somehow).

What's the reason for flushing the buffers, is it really necessary?

jannic:
Another observation: At its current place, that `TCIOFLUSH` seems to be completely useless, as it happens before the port is configured to its final baud rate (and other settings).
Therefore, there is still a window where garbled data could be received and placed in the input buffer.
My suggestion still is: Just remove that line.

kaluk1321:
> Another observation: At its current place, that `TCIOFLUSH` seems to be completely useless, as it happens before the port is configured to its final baud rate (and other settings). Therefore, there is still a window where garbled data could be received and placed in the input buffer. My suggestion still is: Just remove that line.

Damn i was just going to ask a question about how to clear the buffer properly since i was still receiving garbage invoking port.clear(ClearBuffer::All) just after opening the port. Is this also useless? Is there a viable way to properly clear the buffer in this version?

jannic:
@kaluk1321, that sounds like a different issue. If you call `port.clear(ClearBuffer::All)` manually, that's obviously after `open()` has been executed successfully. So your call should remove garbled bytes received before the settings were applied.

PR #40, opened 2022-03-15
Remove useless call to tcflush on open

jannic:
Closes: #37 

This is a draft PR for now, as I didn't have time to test it on real hardware.

### DIFF
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -82,6 +82,10 @@ impl TTYPort {
     /// Ports are opened in exclusive mode by default. If this is undesireable
     /// behavior, use `TTYPort::set_exclusive(false)`.
     ///
+    /// If the port settings differ from the default settings, characters received
+    /// before the new settings become active may be garbled. To remove those
+    /// from the receive buffer, call `TTYPort::clear(ClearBuffer::Input)`.
+    ///
     /// ## Errors
     ///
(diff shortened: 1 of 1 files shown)


ijustlovemath:
I wouldn't call this useless; this is pretty standard for getting the device into a known clean state. Otherwise, you may be dropped into the middle of a protocol, which can be hard to handle.

jannic:
I'm not saying that a flush is useless in general. But at that location it doesn't help, and it is sometimes harmful.

It doesn't help as it is called before the serial port is configured. So there is still a window where data received with mismatched settings would cause garbage in the input buffer, which is not flushed but delivered to the application.

And it may be harmful because it is done after DTR is asserted, so it may throw away actual data. (That's the issue reported  in https://github.com/serialport/serialport-rs/issues/37)

As only the application can know the right time to flush buffers, I don't think it should be done automatically during `open`.
This is why I added a reference to `TTYPort::clear` to the doc comment: It's the applications duty to throw away garbage, if necessary.

jessebraham (approved):
> As only the application can know the right time to flush buffers, I don't think it should be done automatically during `open`.

Yeah, I agree with this. I did check the (not longer maintained) `serial` library and they are not calling `tcflush()` in `open`, so at least that's consistent.

Thanks for this!

PR #41, opened 2022-03-18
Fix bug introduced with #39

jannic:
If opening succeeded, automatic closing on drop must be skipped.


jannic:
Btw, as it's not obvious from the description: this is an important fix, as I basically broke opening a serial port on posix systems with #39. Sorry about that, should have noticed that bug before sending the PR.

jessebraham (approved):
Oh yikes, I had totally missed that too. Thanks for clearing this up!

PR #44, opened 2022-04-21
Request exclusive access when opening a posix serialport by default:

### LINKED ISSUE #26: Lock serial port under Linux
https://gitlab.com/susurrus/serialport-rs/-/issues/109
---
In Linux, use `flock(fd, LOCK_EX | LOCK_NB)` to lock the serial port.

See

* [What is the best practice for locking serial ports and other devices in Linux?](https://stackoverflow.com/questions/30316722/what-is-the-best-practice-for-locking-serial-ports-and-other-devices-in-linux) — StackOverflow.
* [Function `nix::fcntl::flock`](https://docs.rs/nix/0.20.0/nix/fcntl/fn.flock.html)

mlsvrts:
Make the posix implementation of TTYPort::open call tiocexcl, which matches the
documented behavior. Port exclusivity can still be dropped via TTYPort::set_exclusive.

Fixes #26

Signed-off-by: mlsvrts <mlsvrts@protonmail.com>

### DIFF
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -105,6 +105,11 @@ impl TTYPort {
             nix::sys::stat::Mode::empty(),
         )?);
 
+        // Try to claim exclusive access to the port. This is performed even
+        // if the port will later be set as non-exclusive, in order to respect
+        // other applications that may have an exclusive port lock.
+        ioctl::tiocexcl(fd.0)?;
+
         let mut termios = MaybeUninit::uninit();
         nix::errno::Errno::result(unsafe { tcgetattr(fd.0, termios.as_mut_ptr()) })?;
         let mut termios = unsafe { termios.assume_init() };
@@ -156,7 +161,7 @@ impl TTYPort {
         Ok(TTYPort {
             fd: fd.into_raw(),
             timeout: builder.timeout,
-            exclusive: false,
+            exclusive: true,
             port_name: Some(builder.path.clone()),
             #[cfg(any(target_os = "ios", target_os = "macos"))]
             baud_rate: builder.baud_rate,


mlsvrts:
**Discussion**
Note that this is a breaking change to the current default behavior of posix's TTYPort::open.

However, it does match the documented/expected behavior:

https://github.com/serialport/serialport-rs/blob/a84a24539c6c02880096a91076f14a1bae0989c1/src/posix/tty.rs#L84

In the interest of keeping the behavior similar to the current system, I didn't introduce `flock`, or any other locking mechanism. It could be a good idea to re-visit this case on a future major-point revision. 

**Reproducing**
Hardware: USB to Serial converter (FTDI) + external RS232 loopback

Terminal 1:
`cat /dev/ttyUSB0`

Terminal 2:
`cargo run --example transmit /dev/ttyUSB0 9600`

Terminal 3:
`cargo run --example transmit /dev/ttyUSB0 9600`

**Expected Results**
- Prior to this patch, both ports will open and write to the port (data rate is x2 in `cat`)
- After applying patch, terminal 3 will fail with: "Error: Device or resource busy"

jessebraham (approved):
Thanks for this, sorry for the wait! Seems to be working as expected on my end.

src/posix/tty.rs
```diff
@@ -105,6 +105,11 @@ impl TTYPort {
             nix::sys::stat::Mode::empty(),
         )?);
 
+        // Try to claim exclusive access to the port. This is performed even
+        // if the port will later be set as non-exclusive, in order to respect
+        // other applications that may have an exclusive port lock.
+        ioctl::tiocexcl(fd.0)?;
+
```
pauschuu, line 112: This is a problem on custom TTY device implementations that do not support this operation, which makes using this library without a patch on those TTYs impossible.

PR #46, opened 2022-04-21
Upgrade nix to 0.24.1, limit features

rtzoeller:
This reduces serialport's compile time slightly, and removes memoffset as an indirect dependency.


jessebraham (approved):
Great, thanks!

PR #47, opened 2022-04-27
Support reporting usb interface

### LINKED ISSUE #9: Enable differentiation of different devices on a single USB serial port adapter
https://gitlab.com/susurrus/serialport-rs/-/issues/81
---
It would be handy if it were possible to differentiate the different ports in a composite USB Serial adapter in some cross platform way, presumably by adding a field to the UsbPortInfo struct. I've attached an `lsusb -v` output for such a device. There seem to be two possible fields that would make this possible. Either the iInterface field (on line 45 and 115) which I believe is generally a human readable string or the bMasterInterface attribute in the CDCACM union (on line 55 and 125) which is a number.

The second method seems to be used to differentiate them by linux as seen below, but iInterface is available through udev.
```
aled@dufous-/dev: ls -l /dev/serial/by-id/
total 0
lrwxrwxrwx 1 root root 13 Jul 14 19:26 usb-Black_Sphere_Technologies_Black_Magic_Probe_DDD5ACC2-if00 -> ../../ttyACM0
lrwxrwxrwx 1 root root 13 Jul 14 19:26 usb-Black_Sphere_Technologies_Black_Magic_Probe_DDD5ACC2-if02 -> ../../ttyACM1
```

It looks like the library actually [already queries](https://gitlab.com/susurrus/serialport-rs/-/blob/master/src/windows/enumerate.rs#L144) the if number on Windows, although it is not exposed to the user. Funnily enough the example in the comment is exactly my use case, funny how life does that.

[example.txt](/uploads/e558152159de13c1f2e2eb5aed8f869d/example.txt)

If anything I said in regards to anything USB doesn't make any sense, I apologize in advance, USB is still kind of a blackbox to me. I can probably work on a pull request on the posix side of things, but I don't own a Windows box so I have no real way of testing windows stuff

mlsvrts:
USB composite devices have an interface identifier that can
be useful when attempting to distinguish between multiple
ports owned by the same USB device.

This feature adds support for an optional 'interface'
member of `UsbInfo` that will be populated for USB composite
devices. Errors detecting this interface information are
currently conservatively converted into `None`.

Closes #9 

Signed-off-by: mlsvrts mlsvrts@protonmail.com

### DIFF
--- a/examples/list_ports.rs
+++ b/examples/list_ports.rs
@@ -28,7 +28,9 @@ fn main() {
                         );
                         println!(
                             "         Interface: {}",
-                            info.interface.as_ref().map_or("".to_string(), |x| format!("{:02x}", *x))
+                            info.interface
+                                .as_ref()
+                                .map_or("".to_string(), |x| format!("{:02x}", *x))
                         );
                     }
                     SerialPortType::BluetoothPort => {
--- a/examples/list_ports.rs
+++ b/examples/list_ports.rs
@@ -26,6 +26,10 @@ fn main() {
                             "           Product: {}",
                             info.product.as_ref().map_or("", String::as_str)
                         );
+                        println!(
+                            "         Interface: {}",
+                            info.interface.as_ref().map_or("".to_string(), |x| format!("{:02x}", *x))
+                        );
                     }
                     SerialPortType::BluetoothPort => {
                         println!("    Type: Bluetooth");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -634,6 +634,8 @@ pub struct UsbPortInfo {
     pub manufacturer: Option<String>,
     /// Product name (arbitrary string)
     pub product: Option<String>,
+    /// Interface (id number for multiplexed devices)
+    pub interface: Option<u8>,
 }
 
 /// The physical type of a `SerialPort`
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -210,7 +210,13 @@ fn port_type(service: io_object_t) -> SerialPortType {
             serial_number: get_string_property(usb_device, "USB Serial Number"),
             manufacturer: get_string_property(usb_device, "USB Vendor Name"),
             product: get_string_property(usb_device, "USB Product Name"),
-            interface: get_int_property(usb_device, "iInterface", kCFNumberSInt8Type)
+            // Apple developer documentation indicates `bInterfaceNumber` is the supported key for
+            // looking up the composite usb interface id. `idVendor` and `idProduct` are included in the same tables, so
+            // we will lookup the interface number using the same method. See:
+            //
+            // https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_driverkit_transport_usb
+            // https://developer.apple.com/library/archive/documentation/DeviceDrivers/Conceptual/USBBook/USBOverview/USBOverview.html#//apple_ref/doc/uid/TP40002644-BBCEACAJ
+            interface: get_int_property(usb_device, "bInterfaceNumber", kCFNumberSInt8Type)
                 .map(|x| x as u8),
         })
     } else if get_parent_device_by_type(service, bluetooth_device_class_name).is_some() {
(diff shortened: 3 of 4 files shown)


mlsvrts:
I need to run tests against some composite USB hardware, so this is marked as a draft for now.

Note: I don't have the hardware to test this change against MacOs.

mlsvrts:
> This would be very useful for me. Is there anything I can do to help this PR be merged ? I do own a mac and an USB2serial adapter.

@maeln Thanks for the bump! Lints are passing now, and I've added some tests to make the Windows regex a little less frightening, so I think linux/windows are in a good place now.

If you want to run `cargo run --example list_ports` on your system and capture the output, that can at least show that this patch doesn't completely break port enumeration on MacOs. If that works, I can see if I can test against my composite hardware in a virtual machine.

maeln:
@mlsvrts Tested and this is what I got :) :
```
➜  serialport-rs git:(support-reporting-usb-interface) cargo run --example list_ports
    Updating crates.io index
   Compiling libc v0.2.126
   Compiling CoreFoundation-sys v0.1.4
   Compiling autocfg v1.1.0
   Compiling IOKit-sys v0.1.5
   Compiling bitflags v1.3.2
   Compiling cfg-if v1.0.0
   Compiling hashbrown v0.11.2
   Compiling os_str_bytes v6.1.0
   Compiling textwrap v0.15.0
   Compiling termcolor v1.1.3
   Compiling strsim v0.10.0
   Compiling clap_lex v0.2.2
   Compiling indexmap v1.8.2
   Compiling mach v0.1.2
   Compiling mach v0.3.2
   Compiling nix v0.24.1
   Compiling atty v0.2.14
   Compiling clap v3.2.2
   Compiling serialport v4.1.1-alpha.0 (/Users/maeln/Work/serialport-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 11.15s
     Running `target/debug/examples/list_ports`
Found 2 ports:
  /dev/tty.BLTH
    Type: PCI
  /dev/tty.Bluetooth-Incoming-Port
    Type: PCI
➜  serialport-rs git:(support-reporting-usb-interface)
```

Seems to work fine :)

jessebraham (approved):
I think this looks pretty good, thank you for the contribution!

I did try to test this out on my MacBook (x86_64, Big Sur) with an ESP-Prog connected, which has an FT2232HL onboard. I frankly know nothing about composite devices so I'm not sure if this counts as one, but it enumerates two different serial ports so I figured it was worth a shot. While the `list_ports` example ran successfully the `Interface` field is always empty. If this is expected behaviour please let me know. I'm not sure if I have any other devices that would work for testing.

```
λ cargo run --example=list_ports
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/examples/list_ports`
Found 6 ports:
  /dev/cu.Bluetooth-Incoming-Port
    Type: Bluetooth
  /dev/tty.Bluetooth-Incoming-Port
    Type: Bluetooth
  /dev/cu.usbserial-14400
    Type: USB
    VID:0403 PID:6010
     Serial Number: 
      Manufacturer: FTDI
           Product: Dual RS232-HS
         Interface: 
  /dev/tty.usbserial-14400
    Type: USB
    VID:0403 PID:6010
     Serial Number: 
      Manufacturer: FTDI
           Product: Dual RS232-HS
         Interface: 
  /dev/cu.usbserial-14401
    Type: USB
    VID:0403 PID:6010
     Serial Number: 
      Manufacturer: FTDI
           Product: Dual RS232-HS
         Interface: 
  /dev/tty.usbserial-14401
    Type: USB
    VID:0403 PID:6010
     Serial Number: 
      Manufacturer: FTDI
           Product: Dual RS232-HS
         Interface: 
```

mlsvrts:
It looks like those ports are correctly enumerating without interface information, and `serialport-rs` is doing the correct thing by reporting `None`. For comparison, here is what I can report with pyserial for a composite USB device in Windows:

```
COM15
    desc: USB Serial Device (COM15)
    hwid: USB VID:PID=1FC9:A101 SER=500100D20F3861D2 LOCATION=1-11:x.2
COM16
    desc: USB Serial Device (COM16)
    hwid: USB VID:PID=1FC9:A101 SER=500100D20F3861D2 LOCATION=1-11:x.0
```

Note the additional `x.0` and `x.2` strings -- those mark the interface indices. Unfortunately I don't think this is postive proof that the Mac implementation works, but it might be the best we can do until someone comes along with this use-case on Mac.

PR #50, opened 2022-05-04
Added blocking support for posix

eranknafo2001:


sirhcel:
Thank you for this PR! What about some rewording on the proposed interface: As the I/O operations will always be blocking (in the sense of non-blocking I/O) wouldn't be `set_timeout(timeout: Option<Duration>)` more clearly express "no timeout" than `set_blocking(true)`?

And just as a side-note: Shouldn't be no timeout the default instead of 0? This would at least be my expectation as one who grew up on Linux.

sirhcel:
> This would be a breaking change. But couldn't a timeout type with an [auto-conversion](https://users.rust-lang.org/t/auto-convert-t-to-option-t/45115) from `Duration` like
> 
> ```rust
> set_blocking<T: Into<Option<Duration>>>(timeout: T) { ... }
> ```
> 
> buy us out?

Giving it [a try at Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=297ee94e15f40df3052f7538f6f8b344), it looks like it could.

PR #51, opened 2022-05-09
Add serde support behind a feature flag

mikeplus64:
Pretty much does what it says on the tin :). 

The dep I've written is `serde = { version = "1.0", features = ["derive"], optional = true }` which I think is fine, but perhaps more versions of serde should be supported?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ categories = ["hardware-support"]
 [target."cfg(unix)".dependencies]
 bitflags = "1.3.2"
 cfg-if = "1.0.0"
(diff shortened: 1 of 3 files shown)


mikeplus64:
I think I've somehow messed up the git history a bit in my attempt to squash my commits and put them at the "end" of current HEAD. Apologies

jessebraham (approved):
Thanks for this! I think the version of the `serde` dependency is probably fine as it is, this is how I normally have seen it in my experience.

No worries regarding the git stuff, it happens. Should be able to rebase just fine (I hope :grin:).

PR #61, opened 2022-08-03
fix `tiocmget` method

philipr-za:
The method used the `ioctl` return value rather than the reference out argument when constructing the returned bitflags

I noticed the problem when trying to use the `read_ring_indicator` method in a unix project and I was just getting a result of zero. I tested using the raw `ioctl` method and realized the return is actually provided by the reference out argument.


delta1:
Nice find, confirmed that arg is used as an out result 

https://elixir.bootlin.com/linux/v5.13.18/source/drivers/tty/tty_io.c#L2529

jessebraham (approved):
Good catch, thank you for fixing this!

PR #65, opened 2022-10-15
Fix for missing `serial_number` on macOS

pavel-kokolemin:
`CFStringGetCStringPtr` was used incorrectly on macOS in `get_string_property`.

From https://developer.apple.com/documentation/corefoundation/1542133-cfstringgetcstringptr :

> This function either returns the requested pointer immediately, with no memory allocations and no copying, in constant time, or returns NULL. If the latter is the result, call an alternative function such as the [CFStringGetCString(_:_:_:_:)] function to extract the characters.

As the result `serial_number` is some cases was empty. After switching  to `CFStringGetCString` it now always set correctly .

### DIFF
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -164,12 +164,18 @@ fn get_string_property(device_type: io_registry_entry_t, property: &str) -> Opti
             return None;
         }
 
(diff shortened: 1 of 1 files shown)


jessebraham (approved):
Good catch, thank you!

PR #68, opened 2022-11-06
Fix OpenBSD compilation issue caused by disabled ErrorKind "use"

n1000:
Compilation error seen in this block:

```rust
        /// Enumerating serial ports on this platform is not supported
        pub fn available_ports() -> Result<Vec<SerialPortInfo>> {
            Err(Error::new(
                ErrorKind::Unknown,
                "Not implemented for this OS",
            ))
        }
```

### DIFF
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -31,7 +31,8 @@ use crate::UsbPortInfo;
     target_os = "ios",
     all(target_os = "linux", not(target_env = "musl"), feature = "libudev"),
     target_os = "macos",
-    target_os = "netbsd"
+    target_os = "netbsd",
+    target_os = "openbsd",
 ))]
 use crate::{Error, ErrorKind};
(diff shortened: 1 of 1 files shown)


sirhcel:
Thank you very much for finding this issue and your contribution!

sirhcel:
Thank you very much for your contribution! And please excuse our sluggish response time.

The fix builds with the existing targets and I will try to get a CI build for OpenBSD back.

sirhcel:
I learned that OpenBSD is currently just a tier 3 target for Rust and not available through `rustup` yet. So I see no trivial path for adding this target to our CI builds. In favor of getting your fix into the next path release (#70), I'd like to postpone adding CI builds for OpenBSD (with #71 or even a later issue).

Do you know of an OpenBSD CI build for Rust where I could have a look at @n1000?

Thank you very much for spotting this issue. Please keep your eyes open in the meanwhile. ;-)

PR #69, opened 2022-11-07
Add loopback test example

mlsvrts:
This example performs a loopback test using real hardware ports

Additionally, some data will be collected and logged during the test to provide some
rudimentary benchmarking information. When 'split-port' is specified, the serial port will be split into two channels that read/write "simultaneously" from multiple threads.

You can also provide the length (in bytes) of data to test with, and the number of iterations to perform or a list of raw bytes to transmit.

To run this example:

1) `cargo run --example loopback /dev/ttyUSB0`
2) `cargo run --example loopback /dev/ttyUSB0 --split-port`
3) `cargo run --example loopback /dev/ttyUSB0 -i 100 -l 32 -b 9600`
4) `cargo run --example loopback /dev/ttyUSB8 --bytes 222,173,190,239`


jessebraham (changes requested):
Thank you for the PR, sorry for the delay in getting to this!

It looks good to me overall, however in its current state we get a panic when running due to both `baudrate` and `bytes` using `-b` as their short flag; removing `short` from one of these options gets the example to a working state.

Could you please just make that one fix (I removed `short` from  `--bytes` but it probably doesn't matter too much)? After that we can get this merged.

mlsvrts:
> Could you please just make that one fix (I removed `short` from `--bytes` but it probably doesn't matter too much)? After that we can get this merged.

Done!

jessebraham (approved):
Thanks!

PR #72, opened 2022-12-01
Update nix and clap dependencies

g0hl1n:
This PR updates all dependencies to their latest version. Namely this is:

* [nix](https://github.com/nix-rust/nix): 0.25.0 => 0.26
* [clap](https://github.com/clap-rs/clap): 3.1.6 => 4.0

The nix update was without changes to the code.

The clap updated needed some changes because of function deprecations in the [clap v3.2.0](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#320---2022-06-13) and [clap v4.0.0](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28) releases.

Furthermore bump the MSRV to 1.60.0 as this is required by clap v4.0.0 (see https://github.com/clap-rs/clap/pull/4060). https://github.com/serialport/serialport-rs/commit/0a743546848cffd27da9851aea98980b18f05b95 reduced the MSRV because of a Yocto dependency. Currently following rust versions are used in the maintained yocto project versions:

- [master](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=master): v1.65.0
- [langdale](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=langdale): v1.63.0
- [kirkstone](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=kirkstone): v1.59.0
- dunfell: no rust support

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
       disable_tests: true
       extra_packages: libudev-dev
       target: x86_64-unknown-linux-gnu
-      toolchain: "1.46.0"
+      toolchain: "1.56.1"
 
   # --------------------------------------------------------------------------
   # BUILD
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
       disable_tests: true
       extra_packages: libudev-dev
       target: x86_64-unknown-linux-gnu
-      toolchain: "1.56.1"
+      toolchain: "1.60.0"
 
   # --------------------------------------------------------------------------
   # BUILD
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ features = [
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
-clap = "3.1.6"
+clap = "4.0"
 
 [features]
 default = ["libudev"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ categories = ["hardware-support"]
 [target."cfg(unix)".dependencies]
 bitflags = "1.3.2"
 cfg-if = "1.0.0"
-nix = { version = "0.25.0", default-features = false, features = ["fs", "ioctl", "poll", "signal", "term"] }
+nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "poll", "signal", "term"] }
 
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
 libudev = { version = "0.3.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
     "Jesse Braham <jesse@beta7.io>",
 ]
 edition = "2018"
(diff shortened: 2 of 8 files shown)


sirhcel:
Thank you very much for your PR and its awesome background checks!

What's your motivation behind bumping these two dependencies? Foremost _clap_? To my understanding, Kirkstone ist the current long-term-support release of Yocto with planned support until at least April 2024. If there is no strong argument why we should bump our MSRV for this dev dependency I prefer to do it in a way which at least works on Kirkstone as well.

Rust is supported on older Yocto releases through [meta-rust](https://github.com/meta-rust/meta-rust). Did you had a certain Yocto release in mind for 0a7435468 @jessebraham?

Bumping nix to 0.26(.1) would make us requiring Rust 1.56 which would be still okay for Kirkstone.

g0hl1n:
Hi @sirhcel,
thanks for your feedback.

My intention on the nix update is to keep it up2date to avoid duplicated nix versions in my project. So basically this is a pure selfish PR :wink: 

For clap I just wanted to adopt the new major release to keep the gap small (and therefore avoid porting it across multiple major versions in the future, which may be cumbersome).

TBH I totally agree that kirkstone support may be nice due to its LTS nature. Nonetheless if we keep that "requirement" we are stuck with MSRV <=1.59.0 until April 2024 at least (according to https://wiki.yoctoproject.org/wiki/Releases).

Another thing which is/was not entirely clear to me is if it is "best-practice" to increase the MSRV in minor versions? Shouldn't this be reserved for Major versions as we possibly break things here?

g0hl1n:
Friendly ping.

Any feedback @sirhcel @jessebraham?

sirhcel:
> My intention on the nix update is to keep it up2date to avoid duplicated nix versions in my project. So basically this is a pure selfish PR 😉

I'm fine with bumping nix.


> For clap I just wanted to adopt the new major release to keep the gap small (and therefore avoid porting it across multiple major versions in the future, which may be cumbersome).
> 
> TBH I totally agree that kirkstone support may be nice due to its LTS nature. Nonetheless if we keep that "requirement" we are stuck with MSRV <=1.59.0 until April 2024 at least (according to https://wiki.yoctoproject.org/wiki/Releases).

I still don't feel comfortable with bumping clap just for the keeping up with the latest release in face of breaking support for the actual Yocto LTS release. Especially if it is just for the examples. I'm using Kirkstone and plan to do so until the next Yocto LTS release is out. Do you know, of a way of using a different MSRV just for dev builds?

If there is one, I'm fine with bumping it for the examples. If not, I prefer to stick to one compatible with Kirkstone. What about splitting this PR then into bumping nix and bumping clap separately?

What's your opinion on this @jessebraham? 


> Another thing which is/was not entirely clear to me is if it is "best-practice" to increase the MSRV in minor versions? Shouldn't this be reserved for Major versions as we possibly break things here?

This sounds reasonable but to be honest, I don't know the rules yet.

g0hl1n:
> 
> If there is one, I'm fine with bumping it for the examples. If not, I prefer to stick to one compatible with Kirkstone. What about splitting this PR then into bumping nix and bumping clap separately?

ACK. Will to that.

Thanks for the feedback :+1:

PR #75, opened 2022-12-18
Update nix dependency

g0hl1n:
This PR updates [nix](https://github.com/nix-rust/nix) dependency to their latest version (0.25.0 => 0.26).

This update was without changes to the code.

This PR is a follow-up of https://github.com/serialport/serialport-rs/pull/72.

#### MSRV
Furthermore bump the MSRV to 1.56.1 as this is required by nix v0.26.0 (see https://github.com/nix-rust/nix/pull/1792). https://github.com/serialport/serialport-rs/commit/0a743546848cffd27da9851aea98980b18f05b95 reduced the MSRV because of a Yocto dependency. Currently following rust versions are used in the maintained yocto project versions:

- [master](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=master): v1.65.0
- [langdale](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=langdale): v1.63.0
- [kirkstone](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/rust?h=kirkstone): v1.59.0 (current LTS)
- dunfell: no rust support

Therefore this MSRV change (1.46.0 => 1.56.1) shouldn't have any effect on YoctoProject releases.

PR #78, opened 2022-12-20
Update MSRV in Cargo.toml and various doc formatting updates

atouchet:
MSRV was updated in #75.


g0hl1n:
Thanks for the catch. Seems I missed that when splitting up the PRs. Sorry.

eldruin (approved):
Thank you!

PR #87, opened 2023-03-21
Update bitflags to v2.0 and satisfy clippy

g0hl1n:
This PR fixes a clippy warning and updates bitflags to v2.0.

For more details please take a look at the specific commit messages.


eldruin (commented):
Thank you!
Looks good to me except for the version specification for bitflags.
Bitflags 2.0.x versions have [1.56.0 as MSRV](https://github.com/bitflags/bitflags/blob/11640f19a7644f3967631733f33ec87b9f911951/Cargo.toml#L8) but [2.y versions can update the MSRV](https://github.com/bitflags/bitflags#rust-version-support) so we should avoid inadvertedly pulling a bitflags that needs a newer MSRV.
For reference, the current Yocto LTS version (kirkstone) features Rust 1.59.0.

g0hl1n:
Thanks for the feeback! Understood.

So should we go for >=2.0.0 and <2.1 or for the obviously also working, wider >=1.3.0 and <2.1?

This would eliminate the duplicated versions in dev builds.

eldruin:
Since I doubt bitflags would release a 1.4 version, I think it would be fine to have wider range.
I would make it >= 1.3.1, though, because 1.3.0 has been yanked.

g0hl1n:
Updated it as discussed and added a MSRV badge to our README :wink:

PR #88, opened 2023-03-21
github workflows: add cargo-deny ci action

g0hl1n:
Use cargo-deny to add more linting depth.

The configured exceptions and settings are based on the current source tree.
Please review if this is correct.

Exceptions should be tackled in future PRs.


eldruin (approved):
Looks good to me, thank you!

PR #63, opened 2022-10-04
Fix unicode names

mlsvrts:
Switch to unicode Windows APIs:

The default use of ANSI Windows APIs prevents some device names from displaying correctly; this can be resolved by using the Unicode API calls. Closes #62

PR #89, opened 2023-03-27
Windows unicode (`_W`) api for all Win32 functions

Crzyrndm:
Based on #84.
[The new changes](https://github.com/serialport/serialport-rs/pull/89/files/a1c13f5608a15c6107c97ff2445dbe6df06e8611..397a5f72aeda475e9997c9d7304451af815562c8)

> We should probably be using the unicode (`SetupDiClassGuidsFromNameW`) or automatic (`SetupDiClassGuidsFromName`) versions of all windows functions. Otherwise devices from non-english speaking manufacturers or users in different localities may run into issues getting correct device names and information.
> 
> See: https://learn.microsoft.com/en-us/windows/win32/intl/unicode-in-the-windows-api

_Originally posted by @mlsvrts in https://github.com/serialport/serialport-rs/pull/84#discussion_r1148426835_

> In that case, I think updating these in another PR instead of putting additional changes here makes more sense to me!

_Originally posted by @mlsvrts in https://github.com/serialport/serialport-rs/pull/84#discussion_r1148673211_

### DIFF
--- a/src/windows/enumerate.rs
+++ b/src/windows/enumerate.rs
@@ -567,6 +567,26 @@ mod test {
 
     use quickcheck_macros::quickcheck;
 
+    #[test]
+    fn from_utf16_lossy_trimmed_trimming_empty() {
+        assert_eq!("", from_utf16_lossy_trimmed(&[]));
+        assert_eq!("", from_utf16_lossy_trimmed(&[0]));
+    }
+
+    #[test]
+    fn from_utf16_lossy_trimmed_trimming() {
+        let test_str = "Testing";
+        let wtest_str: Vec<u16> = as_utf16(test_str);
+        let wtest_str_trailing = wtest_str
+            .iter()
+            .copied()
+            .chain([0, 0, 0, 0]) // add some null chars
+            .collect::<Vec<_>>();
+        let and_back = from_utf16_lossy_trimmed(&wtest_str_trailing);
+
+        assert_eq!(test_str, and_back);
+    }
+
     // Check that passing some random data to HwidMatches::new() does not cause a panic.
     #[quickcheck]
     fn quickcheck_hwidmatches_new_does_not_panic_from_random_input(hwid: String) -> bool {
@@ -740,18 +760,4 @@ mod test {
         let info = parse_usb_port_info(unicode_serial, None).unwrap();
         assert_eq!(info.serial_number.as_deref(), Some("3854356β"));
     }
-}
-
-#[test]
-fn encoding_trimming_utf16() {
-    let test_str = "Testing";
-    let wtest_str: Vec<u16> = as_utf16(test_str);
-    let wtest_str_trailing = wtest_str
-        .iter()
-        .copied()
-        .chain([0, 0, 0, 0]) // add some null chars
-        .collect::<Vec<_>>();
-    let and_back = from_utf16_lossy_trimmed(&wtest_str_trailing);
-
-    assert_eq!(test_str, and_back);
 }
--- a/src/windows/enumerate.rs
+++ b/src/windows/enumerate.rs
@@ -1,6 +1,7 @@
 use std::collections::HashSet;
 use std::{mem, ptr};
 
+use winapi::ctypes::c_void;
 use winapi::shared::guiddef::*;
 use winapi::shared::minwindef::*;
 use winapi::shared::winerror::*;
@@ -8,7 +9,7 @@ use winapi::um::cfgmgr32::*;
 use winapi::um::cguid::GUID_NULL;
 use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::setupapi::*;
-use winapi::um::winnt::KEY_READ;
+use winapi::um::winnt::{KEY_READ, REG_SZ};
 use winapi::um::winreg::*;
 
 use crate::{Error, ErrorKind, Result, SerialPortInfo, SerialPortType, UsbPortInfo};
@@ -327,6 +328,7 @@ impl PortDevice {
 
     // Retrieves the port name (i.e. COM6) associated with this device.
     pub fn name(&mut self) -> String {
+        // https://learn.microsoft.com/en-us/windows/win32/api/setupapi/nf-setupapi-setupdiopendevregkey
         let hkey = unsafe {
             SetupDiOpenDevRegKey(
                 self.hdi,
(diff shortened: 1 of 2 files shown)


RossSmyth:
I believe this would be a good change, since using the wide function are generally recommended and would make their usage consistent throughout the codebase. Just needs to be rebased.

Crzyrndm:
> I believe this would be a good change, since using the wide function are generally recommended and would make their usage consistent throughout the codebase. Just needs to be rebased.

This has been (squashed and) rebased. I lack the hardware to test this that I had last year though :/

sirhcel:
Thank you very much for your PR @Crzyrndm and for the recent attention @RossSmyth!


> I lack the hardware to test this that I had last year though :/

Did you had a piece of hardware returning actual UTF-16 codepoints?

mlsvrts:
I recommended this change since we saw issues with reporting device serial numbers as fixed by: https://github.com/serialport/serialport-rs/pull/63

It's kind of a case of "if you _can_ do, then someone _will_ do it."

But I doubt it would be a common issue.

src/windows/enumerate.rs
```diff
@@ -401,36 +401,36 @@ fn get_registry_com_ports() -> HashSet<String> {
...
                 // SAFETY: ffi, all inputs are correct
                 let res = unsafe {
-                    RegEnumValueA(
+                    RegEnumValueW(
                         ports_key,
                         idx,
                         val_name_buff.as_mut_ptr(),
                         &mut val_name_size,
                         std::ptr::null_mut(),
                         &mut value_type,
```
sirhcel, line 419: While we are here: Shouldn't we check `value_type` as well? As far as I understand the documentation of [`RegEnumValueW`](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regenumvaluew), the value could (theoretically) be of any type and we are going to process it as wide sting data. And wouldn't [`REG_SZ`](https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-value-types) be the only variant for which we would process correctly?

src/windows/enumerate.rs
```diff
@@ -287,7 +290,7 @@ impl PortDevice {
 
         let mut port_name_buffer = [0u16; MAX_PATH];
         let mut port_name_len = port_name_buffer.len() as DWORD;
-        let value_name: Vec<u16> = "PortName".encode_utf16().chain(Some(0)).collect();
+        let value_name = as_utf16("PortName");
 
         unsafe {
             RegQueryValueExW(
```
sirhcel, line 296: While we are here: Shouldn't we check the type returned through [`RegQueryValueExW`](https://learn.microsoft.com/de-de/windows/win32/api/winreg/nf-winreg-regqueryvalueexw)'s `lpType` as for `RegEnumValueW`?

src/windows/enumerate.rs
```diff
@@ -401,36 +401,36 @@ fn get_registry_com_ports() -> HashSet<String> {
...
                         std::ptr::null_mut(),
                         &mut value_type,
-                        val_data.as_mut_ptr(),
-                        &mut data_size,
+                        val_data.as_mut_ptr() as *mut u8,
+                        &mut byte_len,
                     )
                 };
-                if FAILED(res) || val_data.len() < data_size as usize {
+                if FAILED(res) || val_data.len() < byte_len as usize {
```
sirhcel, line 424: Isn't `val_data`'s length in `u16` units and `byte_len` in `u8`?
```suggestion
                if FAILED(res) || byte_len % 2 != 0 || val_data.len() < byte_len / 2 as usize {
```

src/windows/enumerate.rs
```diff
@@ -287,7 +290,7 @@ impl PortDevice {
 
         let mut port_name_buffer = [0u16; MAX_PATH];
         let mut port_name_len = port_name_buffer.len() as DWORD;
```
sirhcel, line 292: Don't we have to provide the length in terms of `u8` to `RegQueryValueExW` as for `RegEnumValueW` below?
```suggestion
        let mut port_name_len = port_name_buffer.len() * 2 as DWORD;
```

src/windows/enumerate.rs
```diff
@@ -303,9 +306,7 @@ impl PortDevice {
 
         let port_name = &port_name_buffer[0..port_name_len as usize];
```
sirhcel, line 307: Don't we have to check account for the different buffer types `u8` versus `u16` as for `RegEnumValueW` below?

src/windows/enumerate.rs
```diff
@@ -345,8 +346,7 @@ impl PortDevice {
         // Using the unicode version of 'SetupDiGetDeviceRegistryProperty' seems to report the
         // entire mfg registry string. This typically includes some driver information that we should discard.
         // Example string: 'FTDI5.inf,%ftdi%;FTDI'
-        String::from_utf16_lossy(&property_buf)
-            .trim_end_matches(0 as char)
+        from_utf16_lossy_trimmed(&property_buf)
```
sirhcel, line 349: Don't we have to pay attention to buffer versus pointer type and the actually returned type indicated by `SetupDiGetDeviceRegistryPropertyW` as for `RegEnumValueW` as well?

sirhcel (commented):
Looking over the changes there were several points which caught my attention.

src/windows/enumerate.rs
```diff
@@ -401,36 +401,36 @@ fn get_registry_com_ports() -> HashSet<String> {
...
             for idx in 0..num_key_values {
-                let mut val_name_buff = [0i8; MAX_PATH];
+                let mut val_name_buff = [0u16; MAX_PATH];
                 let mut val_name_size = MAX_PATH as u32;
                 let mut value_type = 0;
                 // if 100 chars is not enough for COM<number> something is very wrong
-                let mut val_data = [0; 100];
-                let mut data_size = val_data.len() as u32;
+                let mut val_data = [0u16; 100];
+                let mut byte_len = 2 * val_data.len() as u32; // len doubled
```
sirhcel, line 409: Should't these be checked multiplications because we easily can?

src/windows/enumerate.rs
Crzyrndm: type as `REG_SZ` checks added

src/windows/enumerate.rs
Crzyrndm: fixed

src/windows/enumerate.rs
Crzyrndm: fixed

src/windows/enumerate.rs
Crzyrndm: fixed

src/windows/enumerate.rs
Crzyrndm: I'm unsure as to the motivation for this? There doesn't appear to be any other uses of the checked operations in the lib and the default behaviour is to assert in Debug/test profiles if an overflow is encountered

RossSmyth:
If desired I can give this a more thorough review later in the week. 

> The [API provided by the windows crate](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Registry/fn.RegEnumValueW.html) does't seem to bring more ergonomics witch respect to this buffer handling here.

If it is desired to move to the Windows crate I would recommend that windows-sys is used and not windows-rs ~~because my work computer's anti-virus kills compilation of windows-rs~~ because it is pretty big. Which I do think would be a worthwhile endeavor at some point as winapi is dead.

wrt to buffer handling, the [wide literal macro](https://github.com/microsoft/windows-rs/blob/429666eb724f4a0378974d67adf38c45ce7d78d4/crates/libs/strings/src/literals.rs#L9-L36) could be copied for now to try and reduce allocation if that is important.

src/windows/enumerate.rs
sirhcel: Sorry for the confusion! I did not manage to annotate farther away from your actual change. 😅 I mean the call to `SetupDiGetDeviceRegistryPropertyW` which also provides type information for the returned value and I'm proposing changing this with 727273e.

src/windows/enumerate.rs
sirhcel: My motivation in the first place was to catch overflows. But they won't happen with this code as we are using fixedly sized buffers. So you are right, this is nothing we need to take care of in the short term.

sirhcel:
Rebased to resolve conflict with main branch.

sirhcel:
Thank you for pushing forward with this PR @Crzyrndm! See my replies to https://github.com/serialport/serialport-rs/pull/89#pullrequestreview-2204667053. Resolved the merge conflict with _main_ and added some more cleanup. A local test with `list_ports` looks good.


> If desired I can give this a more thorough review later in the week.

This would be great @RossSmyth.

src/windows/enumerate.rs
Crzyrndm: I think I added a matching check? Not sure if comprehensive though

src/windows/enumerate.rs
```diff
@@ -451,36 +471,40 @@ fn get_registry_com_ports() -> HashSet<String> {
...
-                        val_data.as_mut_ptr(),
-                        &mut data_size,
+                        val_data.as_mut_ptr() as *mut u8,
+                        &mut byte_len,
                     )
                 };
-                if FAILED(res) || val_data.len() < data_size as usize {
+                if FAILED(res)
+                    || value_type != REG_SZ // only valid for text values
+                    || byte_len % 2 != 0 // out byte len should be a multiple of char size
```
RossSmyth, line 496:
```suggestion
                    || byte_len % 2 != 0 // out byte len should be a multiple of u16 size
```
A Rust char is 32 bits and a C char is 8 bits so should be more specific than that.

RossSmyth (approved):
Looks pretty good!

PR #91, opened 2023-04-03
Adds fmt::Debug for dyn SerialPort and related enums.

nathansizemore:
Helpful in circumstances where you want to extract the error from an open and dip out:

```rust
let r = serialport::new().open();
if r.is_err() {
    let e = r.unwrap_err(); // dyn SerialPort does not implement fmt::Debug
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,6 +143,17 @@ pub enum DataBits {
     Eight,
 }
 
+impl fmt::Display for DataBits {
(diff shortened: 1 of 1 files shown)


eldruin:
Sounds like a good idea, thank you!
However, the current version does not build in the MSRV version because the format string capture used is only available in [Rust >= 1.58.0](https://releases.rs/docs/1.58.0/). Could you change that to make it compatible?

eldruin (approved):
Thank you!

PR #98, opened 2023-05-08
Fix memory leaks on macOS

huntc:
There are a number of memory leaks on macOS, particularly when performing the "available_ports" method.

I've added a dependency on `scopeguard` so that we can manage deallocations with a little more sanity. I've also validated that the leaks are repaired via XCode's Instruments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 ### Changed
 ### Fixed
+* A number of memory leaks have been addressed when using serialport-rs on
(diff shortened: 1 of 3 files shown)


sirhcel:
Thank you very much for spotting this issue, providing this fix, and validating it! Looks good to me!

eldruin (commented):
Great work, thank you!
Could you also add an entry to the changelog?

huntc:
> Great work, thank you! Could you also add an entry to the changelog?

Thanks. Commit bcd82377b94c26bb12606945a76be1c867d133c2 addresses the changelog update.

eldruin (approved):
Great work, thank  you!

PR #100, opened 2023-05-10
Maintenance release 4.2.1

sirhcel:
We have integrated several fixes since releasing [4.2.0](https://github.com/serialport/serialport-rs/releases/tag/v4.2.0) but the current main branch breaks semantic versioning due to the field `interface` added to `UsbPortInfo` with 364e1fb957279194a29e74acc435788557182dc2.

There is demand for a maintenance release from our [Matrix channel](https://matrix.to/#/!cnJwoUGZmLziNcGNua:matrix.org/$PbYEq2VsssJlcDt4qdR0UnZAEgMCVBGUH3T7sdg6zFo?via=matrix.org&via=t2bot.io&via=sinic.eu) to release the changes from #98. What about cutting it as everything from _main_ but the breaking changes?

I've checked semantic version compatibility as follows:
```
$ cargo semver-checks check-release --baseline-version 4.2.0
    Updating index
     Parsing serialport v4.2.1-alpha.0 (current)
     Parsing serialport v4.2.0 (baseline, cached)
    Checking serialport v4.2.0 -> v4.2.1-alpha.0 (patch change)
   Completed [   0.013s] 43 checks; 43 passed, 0 unnecessary
```

This PR is for discussing the actual changes and populating the release branch with them. Once we agree that this moves into the right direction, I will do the housekeeping and update version, changelog, ...

PR #101, opened 2023-05-17
Add feature gate for UsbPortInfo.interface (for releasing 4.2.1 from main)

sirhcel:
Adding this field unconditionally is a major breaking change with respect to semantic versioning. Putting it behind a feature gate allows to ship this information early and also still cut 4.x releases from the main branch while still working towards 5.0.

This is an alternative draft to creating (and maintaining) a separate 4.2 release branch as proposed with #100. All maintainers look busy as of now and this approach looks like it allows us to keep a low profile.

I prefer this approach over #100 and if we agree on this way I will prepare the release.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -78,20 +78,38 @@ jobs:
       - name: Build | rust-cache
         uses: Swatinem/rust-cache@v1
 
-      - name: Build | build library
+      - name: Build | build library (default features)
         uses: actions-rs/cargo@v1
         with:
           command: build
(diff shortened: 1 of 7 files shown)


eldruin (commented):
I agree this would ease development for now.
Could you also:
- Add an entry to the changelog?
- Add a note in the Cargo.toml to remove this feature before releasing 5.0 to avoid feature pollution?
- Adapt CI so that the crate is also tested with this feature turned on?

eldruin (approved):
Thank you!

PR #105, opened 2023-05-31
use libudev on musl linux with libudev feature

selfisekai:
udev is a typical way of handling devices on musl-based distributions as well. there is no reason to handle this differently than on gnu libc distros.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -62,7 +62,7 @@ jobs:
           sudo apt-get -qq -y install build-essential curl git pkg-config ${{ inputs.extra_packages }}
 
       - name: Build | install void dependencies
-        if: contains(inputs.container, 'void-linux')
+        if: contains(inputs.image, 'void-linux')
         run: |
           xbps-install -Syu || xbps-install -yu xbps
           xbps-install -yu base-devel bash curl git ${{ inputs.extra_packages }}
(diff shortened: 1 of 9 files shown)


eldruin (commented):
Thank you for this work!
Could you have a look at the CI failures?

eldruin:
It seems the `libudev(-dev)?` library needs to be installed in the CI

selfisekai:
I'm giving up with the ci part.
1. when depending on libudev, there has to be libudev available for the target in pkg-config. using the gnu cross compilers/linkers from debian repositories worked while only musl itself was needed (which is getting provided by rustup). libudev provided by debian works with gnu libc, but not with musl.
2. pkg-config for some reason ignores the env variables pointing to an alpine sysroot with eudev-dev installed, throws a build error that it did not find libudev
3. alpine does not have cross compilers. postmarketos only has cross compilers for aarch64/armv7/armhf. void supports both gnu libc and musl, has cross compilers between anything. void musl does not work on github actions because they bind node.js binaries to the container, and only put musl ones if there is `ID=alpine` in /etc/os-release, so everything has to be a cross to musl compiled on gnu libc. I'm screaming.

eldruin:
I am sorry to hear that. I appreciate your effort and your problem analysis.
However, I am afraid without a way for these changes to be built in the CI we cannot merge this as in the future we would never know if we are shipping broken software.

eldruin:
I will close this now. Please anyone feel free to comment here and take over if you are interested in this.

PR #108, opened 2023-06-16
add space to baud_rate

wxitcode:
add space to baud_rate


sirhcel:
Thank you very much for spotting and fixing that!

PR #111, opened 2023-07-11
No field `path` on type `&SerialPortBuilder`

deid84:
in src/windows/com.rs there are 3 calls to an unexisting struct field called path. Path is an alternative name of the 'port_name' field of SerialPortBuilder structure

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -50,10 +50,10 @@ impl COMPort {
     /// * `InvalidInput` if `port` is not a valid device name.
     /// * `Io` for any other I/O error while opening or initializing the device.
     pub fn open(builder: &SerialPortBuilder) -> Result<COMPort> {
-        let mut name = Vec::<u16>::with_capacity(4 + builder.path.len() + 1);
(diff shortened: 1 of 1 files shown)


deid84:
that's weird... but if i try to read the code in com.rs i cannot find where is declared "path"... in fact i have to manually rename on my laptop in order to proceed and compile. The weird fact is that it started this behavior in last days while i'm using this library since january and it always worked like a charm

sirhcel:
When building the current master [22f854d](https://github.com/serialport/serialport-rs/commit/22f854dbd43966abb6be216db686a06074ccdef1) on Windows 10, everything works like a treat:
```
> git describe --always --dirty
v4.2.2-5-g22f854d
> cargo build
```
How does your repo and build command line look like @deid84? Are the sources you are building from clean?

Building the proposed changes gives me the same errors as in CI.

sirhcel:
Did you find the time to look back into this issue @deid84? Otherwise I would close this issue due to not being able to reproduce it.

deid84:
ok you can close it... i didn't find a way to reproduce the problem

PR #113, opened 2023-07-27
 Fixes the PCI-port bug on Raspberry Pi 4

### LINKED ISSUE #110: UsbPorts are detected as PciPorts on Raspberry Pi 4b
Hey, I know that is a very bad issue, I'm sorry. When I run code like this 

```rust
serialport::available_ports()?
        .into_iter()
        .filter_map(|p| {
            debug!("found device: {p:?}");
            match &p.port_type {
                SerialPortType::UsbPort(info) => {
                    if (info.vid, info.pid) == (VENDOR_ID, PRODUCT_ID) {
                        debug!("It's compatible");
                        return Some(p.port_name);
                    }
                }
                // ...
            }
        }//...
```
on my Laptop, everything works as expected. But when I connect that very same USB device to a Raspberry Pi 4b with a vanilla Raspbian (Bullseye), it is detected with `SerialPortType::PciPort`. This makes it very hard to find the correct device, and requires communication attempts. Coincidentally, all devices I have, use an FTDI chip, so I cannot tell whether this is FTDI specific or not.

KnorrFG:
The Bug is caused by the Fact that the USB controller is connected to
the CPU via PCI, which was not the case on the Raspberry Pi 3.
Because of this, the BUS_ID reported by libudev is PCI. However, if it
is an actual USB device, there will be fields containing the neccessary
information to generate a UsbPortInfo. This patch checks whether all
those fields are available, if the BUS_ID is Pci, and if so uses them to
populate an UsbPortInfo and return a SerialPortType::UsbPort

Fixes #110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 ### Changed
 ### Fixed
+* Fixes a bug on the Raspberry Pi 4, which results in USB-devices being detected as PCI-devices.
+  [#110]
+
 ### Removed
 
 ## [4.2.1] - 2023-05-21
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Changed
 ### Fixed
 * Fixes a bug on the Raspberry Pi 4, which results in USB-devices being detected as PCI-devices.
-  [#110]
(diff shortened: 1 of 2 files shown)


eldruin (commented):
Seems fine to me, thank you!
Could you look into the CI failures?

eldruin (commented):
Thanks! Could you add an entry to the changelog?

eldruin (approved):
Thank you for your work!

KnorrFG:
No problem, I actually need it myself ^^. The sooner it's on crates.io the happier I am :)

Issue #112, opened 2023-07-24
Dependency on unmaintained crates


newAM:
This is minor, but I did not see any existing issue for it.

This crate depends on 3 unmaintained crates:

CoreFoundation-sys: https://github.com/dcuddeback/core-foundation-sys/issues/11
IOKit-sys: https://github.com/dcuddeback/iokit-sys/issues/13
mach: https://github.com/fitzgen/mach/issues/63, [RUSTSEC-2020-0168](https://rustsec.org/advisories/RUSTSEC-2020-0168.html)

```
Crate:     mach
Version:   0.1.2
Warning:   unmaintained
Title:     mach is unmaintained
Date:      2020-07-14
ID:        RUSTSEC-2020-0168
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0168
Dependency tree:
mach 0.1.2
├── IOKit-sys 0.1.5
│   └── serialport 4.2.2-alpha.0
└── CoreFoundation-sys 0.1.4
    ├── serialport 4.2.2-alpha.0
    └── IOKit-sys 0.1.5
```

sirhcel:
Thank you for reporting this issue @newAM! Do you know if there are (de-facto) successors to these crates? Especially to IOKit-sys?

newAM:
For `mach` there is [`mach2`](https://github.com/JohnTitor/mach2/tree/main).
For `IOKit-sys` and `CoreFoundation-sys` the leading alternative is [`apple-sys`](https://crates.io/crates/apple-sys).

sirhcel:
> [...] For `IOKit-sys` and `CoreFoundation-sys` the leading alternative is [`apple-sys`](https://crates.io/crates/apple-sys).

Thank you pointing out these alternatives @newAM! Having a first glance at `apple-sys`, makes it stick out from other sys crates I know by generating is code on the fly where I was expecting a crate just containing the actual bindings.

As far as I'm overseeing this for now, we could experience issues at build time from an incompatible change in the APIs exposed from Apple's SDK but no breaking API from `serialport-rs`. I'm considering the former as having no impact as the APIs in question should be stable. If things get removed from these APIs, then we would have to handle this anyway. On top, we are using IOKit and CoreFoundation only for enumerating ports which also limits potential impact.

As we are not exposing anything from IOKit and CoreFoundation we are safe with respect to semver compatibility when Apple extends APIs with newer Framework versions.

[...]

PR #125, opened 2023-09-17
Swap out unmaintained dependencies for apple-sys

tommy-gilligan:
This deals with #112 

There's probably a way to avoid casting between pointers types.  Any suggestions appreciated.

Not sure why `IORegistryEntryGetParentEntry` needs `plane` argument to be `mut` here. 
https://developer.apple.com/documentation/iokit/1514454-ioregistryentrygetparententry
This is maybe an `apple-sys` issue?
```
-            IORegistryEntryGetParentEntry(device, kIOServiceClass(), parent.as_mut_ptr())
-                != KERN_SUCCESS
+            IORegistryEntryGetParentEntry(
+                device,
+                kIOServiceClass.as_ptr() as *mut i8,
+                parent.as_mut_ptr(),
+            ) != KERN_SUCCESS
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "
 libudev = { version = "0.3.0", optional = true }
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
-apple-sys = { version = "0.2.0", features = ["CoreFoundation", "IOKit"] }
+apple-sys = { git = "https://github.com/tommy-gilligan/apple-sys.git", features = ["CoreFoundation", "IOKit"] }
 mach2 = "0.4.1"
 
 [target."cfg(windows)".dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,7 @@ nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "
 libudev = { version = "0.3.0", optional = true }
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
-CoreFoundation-sys = "0.1.4"
-IOKit-sys = "0.1.5"
+apple-sys = { version = "0.2.0", features = ["CoreFoundation", "IOKit"] }
 mach2 = "0.4.1"
 
 [target."cfg(windows)".dependencies]
--- a/deny.toml
+++ b/deny.toml
@@ -83,6 +83,8 @@ allow = [
     "Apache-2.0",
     "Unicode-DFS-2016",
     "BSD-2-Clause",
+    "BSD-3-Clause",
+    "ISC",
 ]
 # List of explicitly disallowed licenses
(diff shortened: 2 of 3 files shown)


sirhcel:
Thank you for this PR @tommy-gilligan! Noting your last update to it, i've got two questions:
* Isn't IOKit already provided by apple-sys? The feature in [`sys/Cargo.toml`](https://github.com/youknowone/apple-sys/blob/2f0c553a2010be4da0441096347255fe2e720c01/sys/Cargo.toml#L131) and the example from [How to use?](https://github.com/youknowone/apple-sys/tree/2f0c553a2010be4da0441096347255fe2e720c01#how-to-use) suggest this to me.
* The code of apple-sys is licensed under BSD 3 clause. Let's assume that this license also applies to the generated code this should be compatible with the MPL 2.0 used by serialport-rs. The license of the tool apple-bindgen shouldn't affect us here. Is there a way of distinguishing between built-time tools and code used by our crate?

sirhcel:
Sorry for being busy for so long. 

> How was `serialport-rs` ever working on iOS if IOKit wasn't available? Was it successfully linking against macOS IOKit but then not being tested? I'm going to leave this open for now but try a fresh fork exploring other alternatives (DriverKit? Creative use of Posix API? Other Apple crates?)

To be honest, I have never touched this area so far and I have little insight in iOS. May be IOKit was "just there" but not officially supported. Or nobody has built serialport-rs for iOS for a long time and this is the reason why there are no issues with respect to that in our actual tracker.

sirhcel:
> You are asking about how we distinguish between licensing of build-time tools and generated code in `serial-rs` crate metadata?

Exactly. Because I don't feel like changing the licensing of serialport-rs just for getting the generator from apple-sys on board. Call me old-fashioned but I still don't feel comfortable with a dependency where the actual code we are building against only resides in the plumbing of Cargo builds on developer machines.

For this reason, I checked out core-foundation-sys and io-kit-sys in the meantime. I got some required patches merged into io-kit-sys. There are only two things preventing us from using these two crates:

* A new version of core-foundation-sys getting published on creates.io
    * We require a currently upublished function
    * 0.8.5 contains this one but got yanked on crates.io
* An update release of io-kit-sys which depends on the new release of core-foundation-sys

Building a [test version of serialport-rs](https://github.com/serialport/serialport-rs/commit/36da0716a8b5eefa722e2679dad2cd06380579fa) against unreleased code of these creates already works. What is your opinion on this more traditional approach?

sirhcel:
> This is unfortunate. I've definitely faced problems like this elsewhere. Do you have any tips or favourite methods for dealing with stale releases of your dependencies?

I've done a lot of doorbell ringing in the past by creating issues asking for new releases (like https://github.com/servo/core-foundation-rs/issues/645). This includes friendly pings and nagging but worked in all cases so far.

I've cleaned up my test and created #136 from it. So I'm closing this issue in favor of the former. Thank you very much for getting this issue rolling again and the fruitful discussion! :)

PR #127, opened 2023-09-27
Update bitflags dependency

pattop:
Limiting bitflags to <2.1.0 causes compatibility issues and build failures with other crates, e.g:

error: failed to select a version for `bitflags`.
    ... required by package `bindgen v0.68.1`
    ... which satisfies dependency `bindgen = "^0.68.1"` of package ...
versions that meet the requirements `^2.2.1` are: 2.4.0, 2.3.3, 2.3.2, 2.3.1, 2.3.0, 2.2.1

all possible versions conflict with previously selected packages.

  previously selected package `bitflags v2.0.2`
    ... which satisfies dependency `bitflags = ">=1.3.1, <2.1.0"` (locked to 2.0.2) of package `serialport v4.2.2`
    ... which satisfies dependency `serialport = "^4.2.2"` (locked to 4.2.2) of package ...

failed to select a version for `bitflags` which could resolve this conflict

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ keywords = ["serial", "hardware", "system", "RS232"]
 categories = ["hardware-support"]
 
 [target."cfg(unix)".dependencies]
-bitflags = ">=1.3.1, <2.1.0"
+bitflags = "2.4.0"
 cfg-if = "1.0.0"
 nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "poll", "signal", "term"] }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 ### Added
 ### Changed
+
+* Update `bitflags` dependency to 2.4.0.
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ keywords = ["serial", "hardware", "system", "RS232"]
 categories = ["hardware-support"]
 
 [target."cfg(unix)".dependencies]
-bitflags = ">=1.3.1, <2.1.0"
+bitflags = "1.3.1, 2.0.0"
```
eldruin, line 18:
```suggestion
bitflags = "2.4.0"
```

eldruin: The current status does not compile, though.
Just `bitflags = "2.4.0"` should work for everybody. Its MSRV is also 1.56.0.

eldruin (commented):
Thanks for bringing this up!

Cargo.toml
pattop: Updated, thanks.

eldruin (commented):
Thank you!
Sorry, could you also add an entry to the changelog about this update?

pattop:
> Thank you! Sorry, could you also add an entry to the changelog about this update?

I've updated the patch.

eldruin (approved):
Thank you!

PR #130, opened 2023-10-19
open with O_CLOEXEC

mrtnlrsn:
This fixes a problem with Tauri auto update: After updating the tauri application it is restarted, but the serial files were not closed, and therefore the application could not open them again.

With this flag the files are closed when executing `execve()` to start the new application instance.

Change-Id: I102ba33411135a8e89e3203f9dad910d385feeee


eldruin (commented):
Interesting. Could you add an entry to the changelog about this?

eldruin (approved):
Thank you!

PR #131, opened 2023-10-29
Fix serial number reporting on Windows for some devices

xobs:
Some devices have their serial numbers misreported. For example, the BMP device `USB\VID_1D50&PID_6018&MI_02\6&A694CA9&0&0000` gets a reported serial number of `6` rather than `A694CA9`.

Fix this and update the testcase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ project adheres to [Semantic Versioning](https://semver.org/).
 * Fixes a bug on Linux without udev where `available_ports()` returned wrong
   device file paths.
   [#122](https://github.com/serialport/serialport-rs/pull/122)
+* Fixes a bug on Windows where some USB device serial numbers were truncated.
+  [#131](https://github.com/serialport/serialport-rs/pull/131)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


xobs:
Additionally, several clippy lints were fixed as a result of looking over this code.

eldruin (commented):
Nice, thank you!
Could you add a short entry to the changelog as well?

xobs:
I'm actually not sure about the validity of this approach.

image

For example, on my system, the serial number shows up as `3aab607b`. That's because my hwid is `8&3aab607b&0&0000`. However, my actual serial number is `97B6DC14`.

It's possible to get the serial number by trimming `&MI_00` from the port, then looking for a matching `ParentIdPrefix` value that matches the hwid.

Is there still interest in this patch if it returns "a" serial number rather than "the" serial number?

eldruin:
I see. I am not sure myself. Maybe somebody else can weigh in?

danielstuart14:
Yes! Even if it doesn't return "the" serial number, it still is a lot better than before.
With this PR, I can distinguish between two identical devices (each with two CDC endpoints), which I couldn't do before.

Prior:
Screenshot 2023-11-10 164936

After:
Screenshot 2023-11-10 164919

xobs:
Sounds good. I've updated CHANGELOG.md as requested.

eldruin (approved):
Great, thank you!

Issue #134, opened 2023-11-19
Linux: `serialport::available_ports()` returns generic USB strings instead of actual descriptor values


makkarpov:
On Linux, manufacturer and product strings for USB serial ports are populated with generic strings from VID/PID database instead of actual string descriptors as expected.

For example, for a test device with VID `0x1209` (https://pid.codes) and PID `0x0002` (reserved test value) you would get the following (a bit useless) enumeration: 

```
SerialPortInfo {
    port_name: "/dev/ttyACM0",
    port_type: UsbPort(
        UsbPortInfo {
            vid: 4617,
            pid: 2,
            serial_number: None,
            manufacturer: Some(
                "Generic",
            ),
            product: Some(
                "pid.codes Test PID",
            ),
        },
    ),
}
```

However, device itself contains much more detailed string data, which could be used to locate the device much more precisely:

```
[488989.469938] usb 11-4.1.4: new full-speed USB device number 48 using xhci_hcd
[488989.559135] usb 11-4.1.4: New USB device found, idVendor=1209, idProduct=0002, bcdDevice= 2.00
[488989.559143] usb 11-4.1.4: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[488989.559148] usb 11-4.1.4: Product: Your cool product
[488989.559151] usb 11-4.1.4: Manufacturer: Your manufacturer name
```

Alternate way of querying correct Linux device info could be found in pyserial repository: https://github.com/pyserial/pyserial/blob/master/serial/tools/list_ports_linux.py

udev also contains correct strings, but I'm not sure why serialport-rs extracts generic data instead.

sirhcel:
Thank you for bringing up this issue! I can confirm this behavior. We are using [udev for retrieving this information](https://github.com/serialport/serialport-rs/blob/60008114c5393c40ed045a1394dad5569b54345e/src/posix/enumerate.rs#L105C35-L105C58), strange, that it does not report the actual value from `iManufacturer` and `iProduct`.

Thanks for the pointer to pySerial. I will have a look there.

sirhcel:
It looks like we are preferring the information from udevs hardware database over the one provided by the usb device itself. I'm working on #137 to sort this out. Could you please give these changes a try?

Issue #129, opened 2023-10-17
Wrong informations with Com port 


charleselie:
Hello,

I'am building an app on my PC in rust using Slint for the GUI.
This app is communicating through USB with a board with rust embedded code on it. 

On the embedded code I can change inforamtions such as "manufacturer", "serial number", .... 
But unfortunately, I can't get all those informations on my app, for example the manufacturer is always wrong.

Is there a fix for that ? or another lib to use ? 

charles-élie

sirhcel:
Thank you for reporting this issue! 


> On the embedded code I can change inforamtions such as "manufacturer", "serial number", .... But unfortunately, I can't get all those informations on my app, for example the manufacturer is always wrong.

Could you please provide an example of what `dmesg` or `lsusb -v` are reporting and what ` SerialPortInfo` is returned for the port in question? May be the cause is the same as in #134.


> Is there a fix for that ? or another lib to use ?

We have to look into it. There are other serial port abstractions in rust around. But I have not worked with them and cannot tell you off the cuff.

sirhcel:
Hello @charleselie, we've merged #137 which is likely to solve your issue. Could you please give the current master a try? You can use this revision by replacing your dependency with the [unreleased changes](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) in `Cargo.toml`:
```
serialport = { git = "https://github.com/serialport/serialport-rs.git", rev = "e10d3c9" }
```

Feel free to reopen this issue if your problem still exists with.

PR #137, opened 2023-11-20
Prefer manufacturer and model information from device over udev database

sirhcel:
* This addresses #134 and likely #129
* Previously, our lookup preferred the information from udev's hardware database over the information provided by USB devices itself
    * The former is static an will for example provide the following for a SiliconLabs microcontroller eval board with a J-LINK interface
        ```
        $ cargo run --release --example list_ports
        [...]
          /dev/ttyACM0
            Type: USB
            VID:1366 PID:0105
             Serial Number: [...]
              Manufacturer: SEGGER
                   Product: Zeppelin USB 3.0 Host controller
        [...]
        ```
    * Which even looks like faulty product information as this board has an USB 2.0 interface and is definitely no host controller
    * With this PR, the information from the device will be provided
        ```
        $ cargo run --release --example list_ports
        [...]
          /dev/ttyACM0
            Type: USB
            VID:1366 PID:0105
             Serial Number: [...]
              Manufacturer: Silicon Labs
                   Product: J-Link OB
        [...]
        ```
* This change will break existing lookups due to different manufacturer and product information returned
    * We can state this in our changelog
    * Do we have the chance to prevent automatic updates by `cargo update` by releasing this change with a minor release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ project adheres to [Semantic Versioning](https://semver.org/).
   handles when starting a child process. In particular this means that a serial
   device can be reopened after making SW update of a Tauri application.
   [#130](https://github.com/serialport/serialport-rs/pull/130)
+* Prefer USB device manufacturer and model information from the actual USB
+  device over the information from udev's database.
+  [#137](https://github.com/serialport/serialport-rs/pull/137)
+* Raise MSRV from 1.56.1 to 1.59.0.
+  [#137](https://github.com/serialport/serialport-rs/pull/137)
 
 ### Fixed
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -56,7 +56,7 @@ jobs:
       disable_tests: true
       extra_packages: libudev-dev
       target: x86_64-unknown-linux-gnu
-      toolchain: "1.56.1"
+      toolchain: "1.59.0"
 
   # --------------------------------------------------------------------------
   # cargo-deny
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,14 +48,66 @@ jobs:
 
   # --------------------------------------------------------------------------
   # MSRV
+  #
+  # Check at least once per platform.
 
-  msrv:
+  msrv-aarch64-apple-darwin:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      target: aarch64-apple-darwin
+      toolchain: "1.59.0"
+
+  msrv-arm-linux-androideabi:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      target: arm-linux-androideabi
+      toolchain: "1.59.0"
+
+  msrv-x86_64-unknown-freebsd:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      target: x86_64-unknown-freebsd
+      toolchain: "1.59.0"
+
+  msrv-x86_64-unknown-linux-gnu:
     uses: ./.github/workflows/build.yaml
     with:
       disable_extra_builds: true
       disable_tests: true
       extra_packages: libudev-dev
       target: x86_64-unknown-linux-gnu
+      toolchain: "1.59.0"
+
+  msrv-x86_64-unknown-linux-musl:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      extra_packages: gcc-aarch64-linux-gnu
+      target: aarch64-unknown-linux-musl
+      toolchain: "1.59.0"
+
+  msrv-x86_64-pc-windows-msvc:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      runs_on: windows-2019
+      target: x86_64-pc-windows-msvc
+      toolchain: "1.59.0"
+
+  msrv-x86_64-unknown-netbsd:
+    uses: ./.github/workflows/build.yaml
+    with:
+      disable_extra_builds: true
+      disable_tests: true
+      target: x86_64-unknown-netbsd
       toolchain: "1.59.0"
 
   # --------------------------------------------------------------------------
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "
 
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
 libudev = { version = "0.3.0", optional = true }
-unescape = "0.1.0"
+unescaper = "0.1.3"
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 core-foundation-sys = "0.8.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "
 
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
 libudev = { version = "0.3.0", optional = true }
+unescape = "0.1.0"
 
 [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
 core-foundation-sys = "0.8.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,8 +5,8 @@ authors = [
     "Bryant Mairs <bryant@mai.rs>",
     "Jesse Braham <jesse@beta7.io>",
 ]
-edition = "2018"
-rust-version = "1.56.1"
+edition = "2021"
+rust-version = "1.59.0"
 description = "A cross-platform low-level serial port library."
 documentation = "https://docs.rs/serialport"
 repository = "https://github.com/serialport/serialport-rs"
--- a/README.md
+++ b/README.md
@@ -109,7 +109,8 @@ can help debug software or hardware errors.
 
 # Dependencies
 
-Rust versions 1.59.0 and higher are supported.
+Rust versions 1.59.0 and higher are supported by the library itself. There are
+examples requiring newer versions of Rust.
 
 For GNU/Linux `pkg-config` headers are required:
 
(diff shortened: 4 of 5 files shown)


eldruin (commented):
Thank you!
I would see a problem with this released in a patch release but I think it is fine for a minor release. The API does not break in the sense that it continues to compile, it just provides different strings. I would not see that information as SemVer-breaking.
I would like to avoid the `unscape` dependency, though. That crate seems rather abandoned.

sirhcel:
Thanks for your feedback and starting the discussion @eldruin! 

> Thank you! I would see a problem with this released in a patch release but I think it is fine for a minor release. The API does not break in the sense that it continues to compile, it just provides different strings. I would not see that information as SemVer-breaking.

If a minor release prevents automatic updates by `cargo update` than this would be ideal. I don't feel like we should break bespoke routines for finding serialports with such a semiautomatic update. I tried to find a statement with respect to this in the Cargo docs but failed so far. Do you know if this is clearly stated somewhere?

If minor updates would happen automatically by Cargo, I would prefer, postponing this change to the next major release and put it behind a feature gate until then.


> I would like to avoid the `unscape` dependency, though. That crate seems rather abandoned.

It definitely looks like being abandoned. Do you know of a worthy alternative? I've seen [unescape-rs/#8](https://github.com/saghm/unescape-rs/issues/8) recommending [unescaper](https://github.com/hack-ink/unescaper), but it is released only under GPLv3 and therefor not compatible with the licensing of serialport-rs. Started the discussion for dual-licensing _unescape-rs_ with https://github.com/hack-ink/unescaper/issues/26.

On the other hand is the code from _unescape_ sufficiently simple and according to the stats on crates.io it is still well used today. If there is serious trouble with this _unescape_, I would espect it to pop up as a RUSTSEC advisory which will show in CI. Maintaining our own unescape code and tests does not seem worth it to me in the first place. And as _unescape_ as a dependency is used only internally, we could switch to another crate or own implementation at any time.

With that said, how would you prefer to continue with unescaping? I have no strong opinion yet.

AurevoirXavier:
Hi, you can now use https://github.com/hack-ink/unescaper/releases/tag/v0.1.3.

:)

sirhcel:
Thank you very much @AurevoirXavier for your quick response and action!

Now reality kicks in and I see that should have spent some thoughts on the compatibility of Rust versions and Cargo features beforehand. But I just built with the latest release of Rust locally and did not spot this incompatibility in the first place. 🤦 I will think about how we could get this resolved ...

AurevoirXavier:
> I will think about how we could get this resolved ...

It's 2023 now, almost 2024. I really suggest you guys move to at least the edition 2021.

sirhcel:
> It's already 2023, almost 2024. I highly recommend that you all upgrade to at least the 2021 edition.

This is indeed a good point @AurevoirXavier. One (if not the only) reason for sticking to this old MSRV is the support or the latest Yocto LTS (Kirkstone) release for being able to build serialport-rs in this environment. And to disclose my affiliation with it: I'm doing this at my day job and I am therefor opinionated on this. While writing this comment, I saw that the next LTS (Nanbield) is out since some days and ships with Rust 1.70. So a bigger bump is within reach.

But back to business: Our current MSRV already allows using the 2021 edition and I don't know for which reason (if any) we are still using the 2018 edition. If I had to guess I would assume that there was no focus on the edition when bumping MSRV in the past.

To my understanding, the Rust edition is a choice which just affects how serialport-rs is build and is independent of the edition chosen by its dependencies. What actually matters is the Version of Cargo we are using and its feature _named-profiles_ got stabilized with [1.57](https://github.com/rust-lang/cargo/blob/9b13310ca596020a737aaa47daa4ed9ff8898a2f/src/cargo/core/features.rs#L1048).

And now the good news: Yocto Kirkstone's Rust support has been bumped to 1.59.0 this spring and updating our MSRV to this release would buy us out in the sort term. See [this test](https://github.com/serialport/serialport-rs/compare/14ab9c9ab2838530cdb0e3eee02c918dbacec71c...sirhcel:serialport-rs:3dbf8e32ac922ca390790ad2d31ee7e2322f327e) and its [build](https://github.com/sirhcel/serialport-rs/actions/runs/6984584622).

sirhcel:
What's your opinion on bumping our MSRV to 1.59.0 with a minor release @jessebraham and @eldruin. I'm asking because during my testing I figure out that our examples no longer build with 1.56.1 due to an MSRV upgrade from a dev sub-dependency.

Its _serialport_ > _clap_ > _clap_lex_ > _os_str_bytes_ which bumped its MSRV with the release 6.6.0 to 1.60.1 with a minor release. I learned that the resolver automatically increases minor releases for versions 1.0.0 and above (`1.0.0` == `>=1.0.0` and `<2.0.0`). So even a small bump could be a breaking change in some fringe cases.

I also learned today that our MSRV is even behind the Rust release shipped with Debian stable. 😂  So I bet bumping our MSRV to 1.59.0 likely only affects people building our crate in a well aged Yocto environment and I'm inclined to being fine with bumping our MSRV with a minor release.

sirhcel:
> I would like to avoid the `unscape` dependency, though. That crate seems rather abandoned.

Would you be fine with using _unescaper_ @eldruin?

jessebraham:
> What's your opinion on bumping our MSRV to 1.59.0 with a minor release

I think that's reasonable. I had actually bumped this library to `edition = "2021"` previously, however this caused issues for some people (the Yocto folks, I think?) at that time. If this is no longer an issue then there is no reason not to bump the MSRV, IMO.

sirhcel:
Updated MSRV and docs as well. Also extended the MSRV checks in CI to all platforms mentioned in README (with Linux getting some special treatment with respect to udev and Sysfs/musl).

Any objections against merging @eldruin and @jessebraham? And since we've accumulated some fixes since 4.2.2: Any objections with cutting a new release 4.3.0 then?

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ project adheres to [Semantic Versioning](https://semver.org/).
   handles when starting a child process. In particular this means that a serial
   device can be reopened after making SW update of a Tauri application.
   [#130](https://github.com/serialport/serialport-rs/pull/130)
+* Prefers USB device manufacturer and model information from the actual USB
+  device over the information from udev's database. This raises the MSRV from
+  1.56.1 to 1.59.0.
+  [#137](https://github.com/serialport/serialport-rs/pull/137)
```
eldruin, line 21:
```suggestion
  device over the information from udev's database.
  [#137](https://github.com/serialport/serialport-rs/pull/137)
  * MSRV was raised from 1.56.1 to 1.59.0.
```
I would prefer the MSRV bump being listed on top level.

eldruin (commented):
Sorry, I was unexpectedly out for a few days.
Thank you for going through all of this.
Bumping the MSRV to 1.59.0 sounds fine to me since we keep kirkstone working.
Using `unscaper` is fine since the license is now compatible. Thank you both for that!
About whether this crate would be updated by `cargo update` or not in some client code: it depends on _their_ dependency specification. For example, if we release this in 4.3.0, if their `Cargo.toml` says `serialport = "4"` or `serialport = "4.2"`, it will be updated. If it says `serialport = "4.2.0"` it will not be, for example. That is all documented [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html).
I would favor a 4.3.0 release indeed.

CHANGELOG.md
sirhcel: Good point. Looked like burying after reading over it a second time. 😅 Fixed.

sirhcel:
Had a quick and dirty check on our reverse dependencies on crates.io with [this gist](https://gist.github.com/sirhcel/ce2ebde35df64bd8159f7a19275c9a39) and it looks like there are only some pretty old versions of `espflash` using a MSRV below 1.59.0 which are currently downloaded only in homeopathic doses:
```
$ cargo run -- --version-requirement "^4" serialport
[...]
total dependencies: 144
dependencies on ^4: 99

infos for espflash
  Info { name: "espflash", num: "2.1.0", downloads: 14525, msrv: Some("1.70"), dependency_req: Some("^4.2.2") }
  Info { name: "espflash", num: "2.0.1", downloads: 21268, msrv: Some("1.65"), dependency_req: Some("^4.2.1") }
  Info { name: "espflash", num: "2.0.0", downloads: 7245, msrv: Some("1.65"), dependency_req: Some("^4.2.1") }
  Info { name: "espflash", num: "2.0.0-rc.4", downloads: 243, msrv: Some("1.65"), dependency_req: Some("^4.2.1") }
  Info { name: "espflash", num: "2.0.0-rc.3", downloads: 6482, msrv: Some("1.63"), dependency_req: Some("^4.2.0") }
  Info { name: "espflash", num: "2.0.0-rc.2", downloads: 187, msrv: Some("1.63"), dependency_req: Some("^4.2.0") }
  Info { name: "espflash", num: "2.0.0-rc.1", downloads: 240, msrv: Some("1.62"), dependency_req: Some("^4.2.0") }
  Info { name: "espflash", num: "1.7.0", downloads: 25226, msrv: Some("1.59"), dependency_req: Some("^4.2.0") }
  Info { name: "espflash", num: "1.6.0", downloads: 3127, msrv: Some("1.59"), dependency_req: Some("^4.2") }
  Info { name: "espflash", num: "1.5.1", downloads: 1727, msrv: Some("1.59"), dependency_req: Some("^4.1") }
  Info { name: "espflash", num: "1.5.0", downloads: 117, msrv: Some("1.59"), dependency_req: Some("^4.1") }
  Info { name: "espflash", num: "1.4.1", downloads: 724, msrv: Some("1.56"), dependency_req: Some("^4.1") }
  Info { name: "espflash", num: "1.4.0", downloads: 487, msrv: Some("1.56"), dependency_req: Some("^4.1") }
  Info { name: "espflash", num: "1.3.0", downloads: 896, msrv: Some("1.56"), dependency_req: Some("^4.0") }
  Info { name: "espflash", num: "1.2.0", downloads: 925, msrv: Some("1.56"), dependency_req: Some("^4") }
  Info { name: "espflash", num: "1.1.0", downloads: 861, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "1.0.1", downloads: 277, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "1.0.0", downloads: 159, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "0.1.3", downloads: 314, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "0.1.2", downloads: 1000, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "0.1.1", downloads: 193, msrv: None, dependency_req: None }
  Info { name: "espflash", num: "0.1.0", downloads: 163, msrv: None, dependency_req: None }
infos for kaleidoscope-focus
  Info { name: "kaleidoscope-focus", num: "0.1.0", downloads: 317, msrv: Some("1.59.0"), dependency_req: Some("^4.2") }
infos for rawzeo
  Info { name: "rawzeo", num: "0.0.1", downloads: 69, msrv: Some("1.66.1"), dependency_req: Some("^4.2.0") }
```
and there are no reverse dependencies to `espflash` which specify an MSRV. So I will sleep well with releasing this change as 4.3.0. What about cutting a release together with #141 @eldruin and @jessebraham?

eldruin (approved):
Looks good to me, thank you!
Yes, a new release with #141 as well sounds good.

PR #139, opened 2023-11-23
update assert_hex crate to 0.4.1

deid84:
i checked the crate cargo toml and it seems that there are no conflict, they're still at rust 2018 version and the minimum required compiler is older than the minimum required by this library


eldruin (approved):
Thank you!

PR #136, opened 2023-11-20
Use maintained sys crates for CoreFoundation and IOKit

sirhcel:
* This is an alternative to #135 which uses creates where the actual code is in the repository and not generated during build time
* Fixes #112
* This PR drive-by fixes the use of `CFDictionaryGet` and `CFGetTypeID`
    * The former could return a `NULL` pointer which [does not seem to be accepted by `CFGetTypeID`](https://github.com/opensource-apple/CF/blob/3cc41a76b1491f50813e28a4ec09954ffa359e6f/CFRuntime.c#L500) by switching to `CFDictionaryGetIfPresent`
    * Using `CFDictionaryGetIfPresent` also allow us to use the already released versions of core-foundation-sys and io-kit and not wait for upcomming releases (like https://github.com/servo/core-foundation-rs/issues/645)

PR #140, opened 2023-11-24
Remove outdated RUSTSEC exemption for macOS

sirhcel:
This got fixed with #136 but the exemption for the mach crate survived but is no longer needed. Keeping it in place would prevent us from getting notified from CI in case this dependency would be reintroduced by accident.


eldruin (approved):
Thank you!

PR #133, opened 2023-11-15
macOS fixes (#49)

astraw:
Hi, this PR updates the changes made by @berkus in #49 to v4.2.2.

The git history is messy because I simply merged @berkus' original patch through the git history and kept it working. I'm fine to rewrite the history if requested.

The remaining changes with main could be further cleaned up. In particular, the iOS code looks to me like it won't compile, but I have not tested it. As I did this work already I wanted to leave it here in the hopes it could be merged or could help someone.

Issue #49, opened 2022-05-03
Some fixes I had to do to make serialport-rs work on MacOS


berkus:
Here's the [diff](https://github.com/metta-systems/serialport-rs/compare/4ae55697f117592a70295b7e3afadc9ce522bb82..7fec572529ec35b82bd4e3636d897fe2f1c2233f).

I'm not sending this as a PR because it's not a PR-quality code and just a quick patch of things that didn't work, like #22 and disabling IOSSIOSPEED that does not work on MacOS.

Maybe it helps you get some ideas for fixing it properly.

jesjos:
I had a project that failed to behave correctly (reading garbled bytes from a serial device) on macOS. I tried the patched code by @berkus and everything worked just fine.

sirhcel:
Thank you very much for this issue report! I had a first look at this issue by

* Receiving some data on macOS with `receive_data` sent from Minicom on Linux
* Sending some data from macOS with `transmit` to Minicom on Linux

The first did not show any received data at all. The data sent from macOS to Linux showed up in Minicom without any issues. This is not the behavior I would have expected here.

But I did not receive any garbled data. @berkus  and @jesjos could you give please some details on the scenario you are experiencing the problems with? Like communication settings, transmission directions, ...

jesjos:
Thank you for your quick reply, @sirhcel !

I'm sorry I didn't provide a full bug report.

My setup is this:

- Communicating with a serial device that uses a fixed baud rate of 115200
- I set baudrate as expected using the serialport-rs
- I read two bytes, I expect them to be little-endian 9
- Instead, I get something like 0x9c

I can actually get the vanilla serialport code working _if I use some other serial reading implementation to set the baud rate first_. So in my case, I use a java library to read the correct two bytes, then I run my rust code and it works fine. But after restarting my rust code it goes back to the wrong baud rate. I suspect something is fishy with how the library sets baud rates, but my skills aren't enough to figure out exactly what.

I'm running macOS Monterey 12.6 on an Intel mac.

DanielJoyce:
I looked into this long ago and there is another ticket open that says baud rate is ignored.

I remember looking at the code and seeing some of the settings specified in the constructor not making it into the lower level serial settings.

sirhcel:
Thank you very much for your feedback! The last days before Christmas are quite busy and I will look into this afterwards. It looks like there is a need for some testing as well.

berkus:
@sirhcel thanks for looking into it!

[...]

PR #142, opened 2023-11-27
Clear input/output buffers on macOS upon TTY open

astraw:
I don't understand why this works, but it resolves problematic behavior (#49) on macOS. Bisecting history, I found commit
3d6c1b25e3430e45290da752e01da96e81beddc3, which removed flushing the IO buffers for a TTY, was problematic on macOS. However, the test `test_osx_pty_pair` which tests pseudo-ttys with baud_rate 0 on macOS (introduced in b54264b2588879bdca4aeb69c5b91f8a4bf980cb) fails if I simply reintroduced flushing. Therefore, this change now re-introduces flushing the IO buffers upon opening a TTY, but only on macOS and only when baud-rate is non-zero.

Thanks to @berkus for the original work.

This supersedes #133.

### DIFF
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -156,6 +156,11 @@ impl TTYPort {
             ));
         };
 
+        #[cfg(any(target_os = "ios", target_os = "macos"))]
+        if builder.baud_rate > 0 {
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Great, thank you for digging into this and finding a simpler solution.

PR #144, opened 2023-12-18
windows: Exclude disabled ports from `available_ports()`

kawadakk:
Uses [`CM_Get_DevNode_Status`](https://learn.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_get_devnode_status) function to query device status and exclude devices reporting problems, such as being disabled by the user in Device Manager.


sirhcel:
Hiding disabled devices or ones with other problems from automagic enumeration sounds reasonable. Thank you very much for your contribution!

PR #158, opened 2024-02-13
Fix write timeout for COM ports

### LINKED ISSUE #124: Serial write stalls application on some Windows machines
Apart from the issue that on some Windows machines (both my laptop and my PC), `read` just blocks the program, also the `write` function blocks the program on my laptop. It works fine on my PC, but on my Laptop it also doesn't write. I tried setting the timeout to 0, still it stalls.

I tried using the [fork by LukaOber](https://github.com/LukaOber/serialport-rs), and that fork fixes the read function, but not the write function. I have verified that the write function is not just broken in that fork, but also in this one.

By "blocking" or "stalling" I mean, that the write function (and in this repo also the read function, as already discussed in #29 ) never returns, not even an error. The program will just stall, regardless of the set timeout.

nskartoredjo:
Resolves the issue where the COM-port write procedure is blocking, which is caused by a static timeout deceleration equal to zero. In general, you would suspect that the `set_timeout` builder method applies a timeout to both the read and write procedures, as the documentation would suggest.

Fixes #124

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -240,14 +240,14 @@ impl SerialPort for COMPort {
     }
 
     fn set_timeout(&mut self, timeout: Duration) -> Result<()> {
-        let milliseconds = timeout.as_secs() * 1000 + timeout.subsec_nanos() as u64 / 1_000_000;
+        let milliseconds = timeout.as_millis();
 
         let mut timeouts = COMMTIMEOUTS {
             ReadIntervalTimeout: 0,
             ReadTotalTimeoutMultiplier: 0,
             ReadTotalTimeoutConstant: milliseconds as DWORD,
             WriteTotalTimeoutMultiplier: 0,
-            WriteTotalTimeoutConstant: 0,
+            WriteTotalTimeoutConstant: milliseconds as DWORD,
         };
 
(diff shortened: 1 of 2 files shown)


src/windows/com.rs
```diff
@@ -240,14 +240,14 @@ impl SerialPort for COMPort {
     }
 
     fn set_timeout(&mut self, timeout: Duration) -> Result<()> {
-        let milliseconds = timeout.as_secs() * 1000 + timeout.subsec_nanos() as u64 / 1_000_000;
+        let milliseconds = timeout.as_millis();
```
nskartoredjo, line 243: `as_millis` is also available. I'm not aware of any breaking difference, let me know if I should revert it back or not

eldruin (commented):
Great, thank you for finding and fixing this.
Indeed it is already documented that the `set_timeout` function "Sets the timeout for future I/O operations." so I would classify this as a fix and not a change.
Could you add an entry about this in the `Fixed` section of the changelog?
I understand this fixes #124

nskartoredjo:
Sorry for the long wait, I've updated the changelog while trying to keep the current format

eldruin (approved):
Looks good to me, thank you!

PR #159, opened 2024-02-14
Add conversions between DataBits, StopBits and their numeric representation

DNedic:
This implements the `TryFrom` and `From` traits for `u8` and `DataBits`/`StopBits` respectively, can be useful for niche uses.

For instance, I had written an `RFC2217` [server implementation](https://github.com/esp-rs/rfc2217-rs), and the lack of these required implementing [all of these myself](https://github.com/esp-rs/rfc2217-rs/blob/main/src/serialport_conversions.rs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 ### Added
+* Added conversions between `DataBits`, `StopBits` types and their numeric representations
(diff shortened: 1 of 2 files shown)


eldruin (commented):
Looks good to me, thank you!
Could you add an entry to the changelog as well?

DNedic:
@eldruin Thanks for the feedback, PTAL

eldruin (approved):
Looks good to me, thanks!

PR #163, opened 2024-03-02
feat: FlowControl: add FromStr impl

jdswensen:
Allow creating a 'FlowControl' enum from a 'str' type. This is useful when parsing strings from config files or command line arguments.


eldruin (commented):
Seems fine to me, could you add an entry to the changelog?

jdswensen:
Done!

eldruin (approved):
Thank you!

PR #166, opened 2024-03-16
Update macos dependencies

jtakakura:


eldruin:
Using the new dependency versions should be possible just by running `cargo update`. Is there any reason to raise the minimum compatible version like this?

jtakakura:
I misunderstand updating the dependency version.

PR #169, opened 2024-03-27
Fix nightly

RossSmyth:
Currently this crate does not build on nightly due to lint changes. Specially there are two lines that cause compilation failure:

A `deny(unused)` lint:
https://github.com/serialport/serialport-rs/blob/84f6066ffd92056829b91771afc7eebcbc6642e2/src/lib.rs#L23

And importing a trait included in the prelude:
https://github.com/serialport/serialport-rs/blob/84f6066ffd92056829b91771afc7eebcbc6642e2/src/lib.rs#L35

If you attempt to build on nightly you get this error:
```log
error: the item `From` is imported redundantly
   --> src\lib.rs:35:5
    |
35  | use std::convert::From;
    |     ^^^^^^^^^^^^^^^^^^
    |
   ::: C:\Users\rsmyth\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\prelude\mod.rs:129:13
    |
129 |     pub use core::prelude::rust_2021::*;
    |             ------------------------ the item `From` is already defined here
    |
note: the lint level is defined here
   --> src\lib.rs:23:5
    |
23  |     unused
    |     ^^^^^^
    = note: `#[deny(unused_imports)]` implied by `#[deny(unused)]`

error: could not compile `serialport` (lib) due to 1 previous error
```

This fixes the lint, and add nightly to CI. I'm not that great at GitHub actions wrangling so it may be incorrectly formed. We'll see.
rust-lang/rust#121315

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -46,6 +46,8 @@ env:
   CARGO_TERM_COLOR: always
   # Enable cross compilation for `pkg_config`.
   PKG_CONFIG_ALLOW_CROSS: 1
+  # Deny warnings.
+  RUSTFLAGS: -D warnings
 
 jobs:
   build:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,9 @@ on:
       toolchain:
         default: stable
         type: string
+      continue-on-error:
+        default: false
+        type: boolean
 
 env:
   # While we could define these on a per-job basis, there's no harm in simply
@@ -52,6 +55,7 @@ env:
 jobs:
   build:
     runs-on: ${{ inputs.runs_on }}
+    continue-on-error: ${{ inputs.continue-on-error }}
     steps:
       - name: Build | install dependencies
         if: inputs.runs_on == 'ubuntu-latest'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -245,9 +245,22 @@ jobs:
       disable_tests: true
       target: x86_64-unknown-netbsd
 
+  # --------------------------------------------------------------------------
+  # NIGHTLY BUILD
+
+  aarch64-apple-darwin-nightly:
+    uses: ./.github/workflows/build.yaml
+    with:
+      continue-on-error: true
+      disable_tests: true
+      runs_on: macos-latest
+      target: aarch64-apple-darwin
+      toolchain: nightly
+
   x86_64-pc-windows-msvc-nightly:
     uses: ./.github/workflows/build.yaml
     with:
+      continue-on-error: true
       runs_on: windows-2019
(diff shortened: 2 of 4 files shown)


RossSmyth:
I accidentally made my fix branch with the regex-lite branch of my previous PR rather than main.

sirhcel:
Thank you for spotting this and the fix! My gut instinct would not let CI builds fail due to issues from nightly - however it would be nice to still detect them.

Could you please factor out adding additional CI targets into a separate PR? This would allow to get the first one merged quickly and we could create optional build targets in the latter one.

eldruin:
I would rather remove this from the lib.rs and add it as a flag only to the CI (`RUSTFLAGS="-D unused"`) precisely to avoid this. We should notice lint changes in the compiler but this should not make the crate impossible to build for users.

sirhcel:
> I'm not that great at GitHub actions wrangling so it may be incorrectly formed. We'll see.

CI recipes often feel like Russian roulette with all chambers loaded ... Finally got this right this with 97f8fcc.

sirhcel:
> I would rather remove this from the lib.rs and add it as a flag only to the CI (`RUSTFLAGS="-D unused"`) precisely to avoid this. We should notice lint changes in the compiler but this should not make the crate impossible to build for users.

Thank you for suggesting this! This feels like the right thing to do here for me too. Changed this with 8d94b1a:

- Unused imports are no longer denied and generate a warning.
- Warnings will be denied in CI. We are getting stricter here, but I see no point in integrating code generating warnings.

sirhcel:
Marked nightly builds as optional with `continue-on-error` but sadly this does not give the expected visual indication. The [CI run](https://github.com/serialport/serialport-rs/actions/runs/8474997072) is marked green in the actions, this PR is marked as failed with a red `x` and I would have loved to get a different indicator in both cases for "there might be trouble ahead".

I've seen this in GitLab but this seems not to be available on GitHub as of now (see https://github.com/orgs/community/discussions/11592). At least we've marked the intent in the build recipes and I would let it rest in this state for now.

Nightly builds in CI feel somewhat flaky. CI run [355](https://github.com/serialport/serialport-rs/actions/runs/8474637272) and [377](https://github.com/serialport/serialport-rs/actions/runs/8474997072) give different results and I would have expected all three nightly builds to fail. Is this due to jobs getting scheduled on different runners with different nightly releases?

sirhcel:
Cleaned up the remaining issue with nightly from [CI run 357](https://github.com/serialport/serialport-rs/actions/runs/8474997072/) with 7fec3df.

RossSmyth:
Looks good to me. Yeah making it deny in CI I think is a good idea

eldruin (approved):
Looks good to me, thank you!

PR #168, opened 2024-03-27
Use regex-lite instead of Regex

RossSmyth:
Switches to using regex-lite rather than Regex. This improve comile time significantly as Regex is pretty chonky, and the regex is not used in a performance critical (imo) area. 

My computer is a Windows 10 laptop with an i7-12800H. With rustc 1.77.0, compiling the tests & the rlib.

### Regex
```powershell
> rmdir -r -Force target
> Measure-Command { cargo test --no-run -r }
...
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 42
Milliseconds      : 792
Ticks             : 427928612
TotalDays         : 0.00049528774537037
TotalHours        : 0.0118869058888889
TotalMinutes      : 0.713214353333333
TotalSeconds      : 42.7928612
TotalMilliseconds : 42792.8612

> rmdir -r -Force target
> Measure-Command { cargo build -r }
...
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 12
Milliseconds      : 96
Ticks             : 120962426
TotalDays         : 0.00014000280787037
TotalHours        : 0.00336006738888889
TotalMinutes      : 0.201604043333333
TotalSeconds      : 12.0962426
TotalMilliseconds : 12096.2426
```

### Regex-Lite
```powershell
> rmdir -r -Force target
> Measure-Command { cargo test --no-run -r }
...
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 30
Milliseconds      : 539
Ticks             : 305397015
TotalDays         : 0.000353468767361111
TotalHours        : 0.00848325041666667
TotalMinutes      : 0.508995025
TotalSeconds      : 30.5397015
TotalMilliseconds : 30539.7015

> rmdir -r -Force target
> Measure-Command { cargo build -r }
...
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 4
Milliseconds      : 665
Ticks             : 46651654
TotalDays         : 5.39949699074074E-05
TotalHours        : 0.00129587927777778
TotalMinutes      : 0.0777527566666667
TotalSeconds      : 4.6651654
TotalMilliseconds : 4665.1654
```

### Regex
```powershell
> rmdir -r -Force target
> Measure-Command { cargo test --no-run }
...
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 10
Milliseconds      : 84
Ticks             : 100848322
TotalDays         : 0.000116722594907407
TotalHours        : 0.00280134227777778
[...]

PR #172, opened 2024-03-28
Fix some issues pointed out by Clippy in examples and tests

sirhcel:
They surfaced in the CI builds for #168 and look like new catches from the latest Clippy as it is already present in main (https://github.com/serialport/serialport-rs/actions/runs/8470256382). The intentions in the examples an the test code look clears up the ambiguous parts in the code.


eldruin (approved):
Thank you!

Issue #173, opened 2024-03-29
Configurable buffer size


MathiasYde:
I'm making a program that needs to send 16K bytes over a COM port but my program freezes due to the default buffer size on Windows (4096 bytes).

I don't see an implementation in this crate to change the buffer size, is it even possible?

DanielJoyce:
In general the internal OS level buffers are usually a fixed size of 4K and so can't be changed. You should instead be copying the data out into a program buffer.

PR #174, opened 2024-04-02
Implement configurable buffer size on Windows (#173)

MathiasYde:
I've added the implementation to configure the buffer size on Windows.
I don't know Linux well enough to add a proper implementation, so for now I've made it always return an error with the Unsupported error kind - I don't know if it's even possible to configure the buffer size on Linux.


RossSmyth:
Poking around, this does not seem to be something that is possible on Linux. So I would recommend that this is a method on COMPort specifically.

PR #176, opened 2024-04-03
Update to nix-0.28

### LINKED ISSUE #175: Update dependency on nix to 0.28
The current dependency on nix 0.26 results in multiple crate versions for bitflags.
The latest release of nix (0.28) fixes this.

neumann-paulmann:
While at it, use std::os::fd::owned::OwnedFd instead of custom implementation.

Fixes #175

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ categories = ["hardware-support"]
 [target."cfg(unix)".dependencies]
 bitflags = "2.4.0"
 cfg-if = "1.0.0"
-nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "poll", "signal", "term"] }
+nix = { version = "0.28", default-features = false, features = ["fs", "ioctl", "poll", "signal", "term"] }
 
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
 libudev = { version = "0.3.0", optional = true }
--- a/src/posix/poll.rs
+++ b/src/posix/poll.rs
@@ -1,7 +1,7 @@
 #![allow(non_camel_case_types, dead_code)]
 
 use std::io;
-use std::os::unix::io::RawFd;
+use std::os::fd::AsFd;
 use std::slice;
 use std::time::Duration;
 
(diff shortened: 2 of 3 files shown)


eldruin (commented):
Thank you for your work!
Could you look into the CI failures?
Also, could you add an entry to the changelog about this?

eldruin:
Hmm, this does not seem MSRV-related to me, though. Here is a run with 1.76.0:
https://github.com/serialport/serialport-rs/actions/runs/8535157043/job/23396028026?pr=176#step:7:42

The MSRV is set to 1.59.0, where have you found the 1.36.0 reference?
Changing the MSRV has implications related to the supported platforms.
What do you think @sirhcel ?

sirhcel:
> Hmm, this does not seem MSRV-related to me, though. [...]

No, this looks like an issue with the signedness of `milliseconds` in the explicit conversion from `Duration`. Both target types for the poll timeout support directly converting from `Duration` and we can work around it by using these direct conversions.

Ended up with bf7050c while looking into this. So there is just raising the MSRV left.


> Changing the MSRV has implications related to the supported platforms. What do you think @sirhcel?

At a first glance at least when it comes to the Rust support shipped with Yocto. Dependents using older Yocto releases could still pin their dependency on us to a minor version and we could attempt to scan all publicly visible dependencies. And on the other hand there is also meta-rust-bin which provides the brand spanking new releases of Rust. Let me think and sleep over it.

eldruin:
Did you have time to think about this @sirhcel ?

sirhcel:
I finally managed to. Just to get my perspective on this right: Is bumping our nix dependency tied to switching to `OwnedFd`? Or is the latter a drive-by catch happening in the context of bumping nix?

I'm fine with switching to `OwnedFd` with a minor release as this does not look like a breaking change to me as far as I'm overseeing this as of now.

When it comes to bumping nix, I would postpone this to the next major release as this could be a breaking change for dependents still relying on our MSRV being 1.59.0. I know this is subject to general discussion but we've got our MSRV for dev builds broken recently with our dependencies `serialport` -> `clap` -> `os_str_bytes` (see #186). I would like to spare dependents with tighter MSRV requirements this experience.

max-heller:
Given the adoption of `OwnedFd`, could a `std::os:fd::AsFd` implementation be added as well?

JonathonReinhart:
Hi folks, what's the latest on this effort?

I'm attempting to pull this into a project that's locked-in on `nix` 0.28.0 so this is blocking me from using `serialport`.

Is there anything I can do to help move this forward?

inferiorhumanorgans:
Has there been any progress on updating this crate to use a newer `nix`?  I'm running into an issue where the custom `Drop` impl is causing problems with the borrow checker.

paolobarbolini:
I've made a new attempt at the `nix` upgrade in #263.

sirhcel:
And this work is finally on its way of being integrated into the next major release with #355. This will still take some time to get this over the finish line, but at least the sails are set now.

Thank you very much for getting this started @neumann-paulmann!

PR #181, opened 2024-05-05
Clean up integration tests

sirhcel:
This PRs cleans up some warnings from the doctests and makes tests using serial interfaces putting their cards on the table: The results of the operations under tests were just ignored beforehand and will be checked now.

Having the results checked now, requires the tests to use serial devices designated for testing. To keep a low profile, this is done by passing the devices via environment variables and rests requiring actual serial hardware can be ignored by default with the feature _ignore-hardware-tests_ which is done in CI.


eldruin (approved):
Nice work, thank you!

PR #184, opened 2024-05-06
Streamline integration tests and add first test for read timeouts

sirhcel:
This PR cleans up the following points:

* No longer requiring the the dummy environment variables from #181 (they are evaluated at runtime on demand now)
* Running tests for the `SerialPort` trait on all platforms

It also adds the first test for read timeouts showing the blocking behavior for a timeout of zero as described in #79 and paves the way for integrating #79.


eldruin (approved):
Great, thank you!

PR #185, opened 2024-05-07
feat: allow blocking read

### LINKED ISSUE #12: `read` is always non-blocking even when no data is available to read
https://gitlab.com/susurrus/serialport-rs/-/issues/89
---
The current behavior of `read` on an open serial port is that it always returns immediately, even when there is no data available in the input buffer. This may be in contradiction with the understanding [here](https://gitlab.com/susurrus/serialport-rs/-/blob/master/src/posix/tty.rs#L108) (from tty.rs).

It's also not consistent and not obvious how to make it consistent with the description from https://linux.die.net/man/3/cfmakeraw related to canonical/non-canonical modes. Please note that according to that explanation, raw mode does not guarantee that reads blocks if no data is available. Non-canonical mode with MIN > 0; TIME == 0 or MIN == 0; TIME > 0 or MIN > 0; TIME > 0.

Please clarify what is the intention of serialport-rs implementation, and how to make read block when no data is available for a configurable maximum timeout.

ARizzo35:
Fixes #12, similar to #50 but without the extra functions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -326,7 +326,7 @@ pub struct SerialPortBuilder {
     /// Number of bits to use to signal the end of a character
     stop_bits: StopBits,
     /// Amount of time to wait to receive data before timing out
-    timeout: Duration,
+    timeout: Option<Duration>,
 }
 
 impl SerialPortBuilder {
@@ -375,7 +375,7 @@ impl SerialPortBuilder {
     /// Set the amount of time to wait to receive data before timing out
     #[must_use]
     pub fn timeout(mut self, timeout: Duration) -> Self {
-        self.timeout = timeout;
+        self.timeout = Some(timeout);
         self
     }
 
@@ -459,7 +459,7 @@ pub trait SerialPort: Send + io::Read + io::Write {
     fn stop_bits(&self) -> Result<StopBits>;
 
     /// Returns the current timeout.
-    fn timeout(&self) -> Duration;
+    fn timeout(&self) -> Option<Duration>;
 
     // Port settings setters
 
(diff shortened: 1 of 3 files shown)


sirhcel:
Thank you for the PR! 

> Fixes #12, similar to #50 but without the extra functions.

Yes, it does not need extra functions. But on the other hand switching from `Duration` to `Option<Duration>` breaks the current API and this change needs to wait for the next major release.

What about designating `Duration::MAX` as _no timeout_ as proposed in https://github.com/serialport/serialport-rs/issues/12#issuecomment-1104694319 for the public interface meanwhile?

* Documenting this quirk
* Adding a `no_timeout` function to `SerialPortBuilder` and its implementation for making the intent explicit
* Returning `Duration::MAX` in case no timeout is set

And with the next release switching switching from `Duration` to `Option<Duration>`.

sirhcel:
> I am trying to add support for blocking read in `poll()` / `ppoll()`.

Thank you for the context! So you are trying to read the available data after being notified? Are you switching the serial port to non-blocking for this task? If yes, using `into_raw_fd`?

> Setting a large duration with the current API achieves the same effect, but it is not the actual correct solution.

Yes, the current default timeout is - lets say - somewhat special. I would have expected this to be blocking/no timeout on POSIX systems.


> It's also confusing to use the default timeout of 0 and then see your program using 100% CPU because it is busy looping on a non-blocking read somewhere. Any other Rust library implementation that impls Read will block (e.g. stdin) so things like `for line in lines()` from BufReader do not actually end up consuming a lot of CPU usage, however, when using that pattern with this serial port library, it causes very high CPU usage since the read does not actually block.

Yes, this is another annoyance with the current default.

However, changing the default right now will likely break the somewhat special but expected behavior for other users. Therefor I would like to split things up: Allowing to specify no timeout now more explicitly in the short term. Switching the default behavior with the next major release and provide a cleaned up API for setting timeouts with this release.

eldruin:
> Allowing to specify no timeout now more explicitly in the short term. Switching the default behavior with the next major release and provide a cleaned up API for setting timeouts with this release.

Sounds good to me!

PR #186, opened 2024-05-09
Pin sub-dependecy of clap to be compatible with MSRV

sirhcel:
Fix building from a fresh start with our current MSRV 1.59.0. There has been an incompatible change with respect to the MSRV in _os_str_bytes_ with the minor version update to 6.6.0.

As of 83fd7fb this fails with:
```
$ rm -r target Cargo.lock
$ export SERIALPORT_RS_TEST_PORT_1=DUMMY_1
$ export SERIALPORT_RS_TEST_PORT_2=DUMMY_2
cargo +1.59.0 test --features ignore-hardware-tests
error: package `os_str_bytes v6.6.1` cannot be built because it requires rustc 1.61.0 or newer, while the currently active rustc version is 1.59.0
```

This PR adds a direct dependency to _os_str_bytes_ for tricking the dependency resolution into using a compatible version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,11 @@ serde = { version = "1.0", features = ["derive"], optional = true }
 assert_hex = "0.4.1"
 clap = { version = "3.1.6", features = ["derive"] }
 envconfig = "0.10.0"
+# TODO: Remove pinning this subdependency of clap when we are bumping our MSRV.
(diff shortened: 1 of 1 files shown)


eldruin:
Hmm, sorry I did not look at this soon enough. There is a conflict now in the Cargo.toml. Could you have a look?

sirhcel:
Thank you for pointing it out! This was a quick fix.

PR #188, opened 2024-05-12
Fix intermittent failures with CI build and clean up

sirhcel:
There were recently a lot of hiccups with installing rustup for Windows in CI. For example [here](https://github.com/serialport/serialport-rs/actions/runs/9051912721/attempts/4). This PR switches to a current and actively maintained action for installing the Rust toolchain.

The previous issues were intermittent and could be overcome by trying several times. But let's have a less dicey way for running CI checks. This PR also includes cleaning up warnings from issues with outdated configuration and actions.


eldruin (approved):
Thank you!

PR #189, opened 2024-05-15
Add Mark and Space variants to Parity enum

### LINKED ISSUE #167: Can Add Mark and Space to Parity?
Like Qt
https://doc.qt.io/qt-6/qserialport.html#Parity-enum

QSerialPort::NoParity
QSerialPort::EvenParity
QSerialPort::OddParity
QSerialPort::SpaceParity
QSerialPort::MarkParity

YgorSouza:
Closes #167


YgorSouza:
Note that since the `Parity` enum is not marked `#[non_exhaustive]`, this is a breaking change and requires a major version bump.

Also, I have not tested this yet, and I'm not sure I'll have access to a serial device with mark/space parity to test soon. If someone else could test, that would be appreciated.

src/posix/termios.rs
```diff
@@ -163,6 +163,20 @@ pub(crate) fn set_parity(termios: &mut Termios, parity: Parity) {
             termios.c_iflag |= libc::INPCK;
             termios.c_iflag &= !libc::IGNPAR;
         }
+        Parity::Mark => {
+            termios.c_cflag |= libc::PARODD;
+            termios.c_cflag |= libc::PARENB;
+            termios.c_iflag |= libc::INPCK;
+            termios.c_iflag &= !libc::IGNPAR;
+            termios.c_iflag |= libc::CMSPAR;
```
YgorSouza, line 171: So apparently this only works on Linux. What should it do on other targets? Do the same as odd/even, panic, or something else?

sirhcel: Thank you for starting this PR! Thank you especially for looking at Windows as well! :smile:

Mark and space parity is not supported on all platforms of our POSIX backend. As far as i've looked into it, its controlled by the flag `CMSPAR` where present. I would provide support for mark and space parity only on these ones and make `Parity::Mark` and `Parity::Space` conditionally available via `#[cfg(...)]`.

src/posix/termios.rs
```diff
@@ -163,6 +163,26 @@ pub(crate) fn set_parity(termios: &mut Termios, parity: Parity) {
             termios.c_iflag |= libc::INPCK;
             termios.c_iflag &= !libc::IGNPAR;
         }
+        Parity::Mark => {
+            termios.c_cflag |= libc::PARODD;
+            termios.c_cflag |= libc::PARENB;
+            termios.c_iflag |= libc::INPCK;
+            termios.c_iflag &= !libc::IGNPAR;
+            #[cfg(any(target_os = "linux", target_os = "android"))]
```
sirhcel, line 171: Wouldn't this just silently switch to normal odd even parity checking other backends?

src/posix/termios.rs
```diff
@@ -163,6 +163,26 @@ pub(crate) fn set_parity(termios: &mut Termios, parity: Parity) {
...
+            {
+                termios.c_iflag |= libc::CMSPAR;
+            }
+        }
+        Parity::Space => {
+            termios.c_cflag &= !libc::PARODD;
+            termios.c_cflag |= libc::PARENB;
+            termios.c_iflag |= libc::INPCK;
+            termios.c_iflag &= !libc::IGNPAR;
+            #[cfg(any(target_os = "linux", target_os = "android"))]
```
sirhcel, line 181: Same as already asked above: Wouldn't this just silently switch to normal odd even parity checking other backends?

sirhcel (changes requested):
I would provide `Parity::Mark` and `Parity::Space` only for backends actually supporting it.

src/posix/termios.rs
YgorSouza: I've made the change, but I think it's problematic for a few reasons:

1. If someone wanted to match on the `Parity` (e.g., to provide an icon, color, localized string etc for each variant), they would have to add the same verbose `cfg` everywhere, and they would have to look at the library source to get it right.
2. If another existing target gets support for those variants in the future in one way or another, that would be a new breaking change for the library. Unless the enum is marked `#[non_exhaustive]`, but then you'd have to add a wildcard arm to every match statement, which is annoying and makes the code more brittle.

So maybe it would be better to just make the variants available unconditionally, and document which targets support them. I'm not sure how other libraries like QSerialPort and pySerial handle this. I can try to look into it.

src/posix/termios.rs
YgorSouza: Yes. I'm not sure what is the best way to handle that. Making the whole enum variant conditionally available has some problems as I've noted above. Another option could be to panic if it is not supported, which is arguably better than failing silently.

src/posix/termios.rs
YgorSouza: Maybe there could be a `supported()` method for the enum, returning false for targets which are not supported, so the user can just filter them out without getting into the details of all the Unix variants. And then the `SerialPortBuilder` could panic when the variant is not supported, instead of invoking unspecified behavior.

YgorSouza: Actually, it doesn't need to panic at all. The methods already return `Result`, so I can just make `termios::set_parity` return an error with a descriptive message when the platform is not supported, and propagate that to the public methods. The `open` method already does that:

https://github.com/serialport/serialport-rs/blob/c3209197dfbc0c760c843c7905d49b83b99714e3/src/lib.rs#L385-L397

Does this seem reasonable? I'm going to revert the previous commit and try that to see how it looks.

sirhcel: Thank you very much for pushing this forward! Well, I think, that I now see what you meant with `supported()`. It's `Parity::supported(&self) -> bool`, right. And at the end this might be the Pareto solution to this issue.

Parity is likely being set by an application, it will be done in few cases and even fewer people will use mark and space parity at all. Compared to `Parity` being `#[non_exhaustive]` and having `Mark` and `Space` conditionally available, it has its strength in a less cluttered API.

Let me sleep over it.

sirhcel (approved):

PR #191, opened 2024-05-24
Fix output from example receive_data

sirhcel:
There was no visible output without explicitly flushing stdout on on macOS.


eldruin (approved):
Looks good to me, thanks!

Issue #7, opened 2022-02-07
`TTYPort::read` doesn't respect timeout


jessebraham:
This issue was migrated from GitLab. The original issue can be found here:
https://gitlab.com/susurrus/serialport-rs/-/issues/74
---
I have an application that sometimes just hangs, seemingly forever. I've tracked the problem down to `TTYPort::read`.

I'm setting a timeout on the port and I know this works, as the application experiences timeouts quite regularly. Despite this, the call to `read` doesn't return. I've added some debug output, and tracked the problem to the second call in `read`. [`super::poll::wait_read_fd`](https://gitlab.com/susurrus/serialport-rs/-/blob/3.3.0/src/posix/tty.rs#L459) returns, [`nix::unistd::read`](https://gitlab.com/susurrus/serialport-rs/-/blob/3.3.0/src/posix/tty.rs#L463) doesn't.

The serial port I'm using is connected through USB to a microcontroller, and something really weird is going on over there (longstanding USB-related bug that I wasn't able to track down so far). I'm not surprised that this is causing weird behavior on the host side. However, it would be great, if serialport-rs could respect the timeout setting under such circumstances, if at all possible.

PR #192, opened 2024-05-24
Fix hang on TTY read

danielstuart14:
In some cases, when a underlying error happens to a TTY port between poll::wait_read_fd and unistd::read, the read function would hang waiting for some data that is never received.

This commit sets the port to non-canonical mode, with VMIN = VTIME = 0. With this change, it has the effect of making reads non-blocking, returning right away.

The timeout behaviour is maintained, as prior to reading we call unix::poll through poll::wait_read_fd.

Fixes: https://github.com/serialport/serialport-rs/issues/7

### DIFF
--- a/src/posix/termios.rs
+++ b/src/posix/termios.rs
@@ -5,6 +5,7 @@ use crate::{DataBits, FlowControl, Parity, Result, StopBits};
 use nix::libc;
 
 use std::os::unix::prelude::*;
+use std::time::Duration;
 
 cfg_if! {
     if #[cfg(any(
@@ -181,6 +182,12 @@ pub(crate) fn set_flow_control(termios: &mut Termios, flow_control: FlowControl)
             termios.c_cflag |= libc::CRTSCTS;
         }
     };
+}
+
+pub(crate) fn set_timeout(termios: &mut Termios, timeout: Duration) {
+    let timeout = u128::min(timeout.as_millis() / 100, u8::MAX as u128) as u8;
+    termios.c_cc[libc::VMIN as usize] = 0;
+    termios.c_cc[libc::VTIME as usize] = timeout;
 }
 
 pub(crate) fn set_data_bits(termios: &mut Termios, data_bits: DataBits) {
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -134,7 +134,9 @@ impl TTYPort {
         nix::errno::Errno::result(unsafe { tcgetattr(fd.0, termios.as_mut_ptr()) })?;
         let mut termios = unsafe { termios.assume_init() };
 
-        // setup TTY for binary serial port access
+        // Setup TTY for binary serial port access
+        // Enable non-canonical mode
+        termios.c_lflag &= !libc::ICANON;
         // Enable reading from the port and ignore all modem control lines
         termios.c_cflag |= libc::CREAD | libc::CLOCAL;
         // Enable raw mode which disables any implicit processing of the input or output data streams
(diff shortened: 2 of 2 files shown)


danielstuart14:
@sirhcel  @eldruin let me know if there is anything you'd like me to change.

src/posix/termios.rs
```diff
@@ -183,6 +184,12 @@ pub(crate) fn set_flow_control(termios: &mut Termios, flow_control: FlowControl)
     };
 }
 
+pub(crate) fn set_timeout(termios: &mut Termios, timeout: Duration) {
+    let timeout = u128::min(timeout.as_millis() / 100, u8::MAX as u128) as u8;
+    termios.c_cc[libc::VMIN as usize] = 0;
+    termios.c_cc[libc::VTIME as usize] = timeout;
```
sirhcel, line 190: The description (and the documentation of termios as well) mention setting `VTIME` to 0 as well. What's the reason for setting it to `timeout` here?

sirhcel (commented):
Thank you for looking into the dusty corners of our issue list @danielstuart14 and creating this PR!


> In some cases, when a underlying error happens to a TTY port between poll::wait_read_fd and unistd::read, the read function would hang waiting for some data that is never received.

Yes, at a first glance this looks like something the actual code does not consider. Do you have an example to reproduce such a situation?


> This commit sets the port to non-canonical mode, with VMIN = VTIME = 0. With this change, it has the effect of making reads non-blocking, returning right away.

Setting `VMIN` and `VTIME` to zero looks like the right thing for not blocking on the read after successfully polling for new data to read. But as annotated below - is `VTIME` set to `timeout unintentionally then?

And isn't there some explicit handling required in case reading returns 0, eventually with `errno` set?


> The timeout behaviour is maintained, as prior to reading we call unix::poll through poll::wait_read_fd.

Wouldn't `VTIME == timeout` change the timeout behavior in when read is callen after polling?

src/posix/termios.rs
danielstuart14: I made this utils function to bridge how timeouts are set on termios (by setting VTIME to a value with 100ms steps) with the Duration type on rust.
If you check at https://github.com/serialport/serialport-rs/blob/896136584a181581b04ef90279a5bfbd2cf50392/src/posix/tty.rs#L181, this function is being called with a duration = 0, which means that VTIME is set to zero.

danielstuart14:
> Thank you for looking into the dusty corners of our issue list @danielstuart14 and creating this PR!
> 
> > In some cases, when a underlying error happens to a TTY port between poll::wait_read_fd and unistd::read, the read function would hang waiting for some data that is never received.
> 
> Yes, at a first glance this looks like something the actual code does not consider. Do you have an example to reproduce such a situation?
> 
> > This commit sets the port to non-canonical mode, with VMIN = VTIME = 0. With this change, it has the effect of making reads non-blocking, returning right away.
> 
> Setting `VMIN` and `VTIME` to zero looks like the right thing for not blocking on the read after successfully polling for new data to read. But as annotated below - is `VTIME` set to `timeout unintentionally then?
> 
> And isn't there some explicit handling required in case reading returns 0, eventually with `errno` set?
> 
> > The timeout behaviour is maintained, as prior to reading we call unix::poll through poll::wait_read_fd.
> 
> Wouldn't `VTIME == timeout` change the timeout behavior in when read is callen after polling?

Hey there!
I'm maintaining a testing jig with 25 serial devices connected to a single computer, and I saw this issue happen maybe three times in the last 6 months. I can't reproduce it on the bench, as the timings (and maybe even the intrinsics of it) are very particular. My guess right now, is that a packet dropped by some issue with the USB controller (which I verified was the case through some dmesg errors) triggered this behavior. Just disconnecting the USB doesn't cause it. I added some debug messages to the code and verified that it was indeed getting stuck in the read.

About VTIME, I'm setting it to zero through the bridge function I created, check my reply above.

Any internal errors should be returned by the unistd "read" implementation. What maybe we could do is return an error if the poll function returned successfully but the read didn't return any data, but it would need some thinking regarding the behavior when timeout = 0 or buf length = 0.

im-0:
This breaks `tokio-serial`. Fix: https://github.com/berkowski/mio-serial/pull/53.

Someone will need to coordinate this with the maintainer of `mio-serial`/`tokio-serial` if you decide to merge this pull request.

PR #193, opened 2024-05-26
Fix USB interface on macOS

parasyte:
When the `usbportinfo-interface` feature is enabled on macOS, the `UsbPortInfo::interface` is always `None`. A careful reading of https://developer.apple.com/library/archive/documentation/DeviceDrivers/Conceptual/USBBook/USBOverview/USBOverview.html#//apple_ref/doc/uid/TP40002644-BBIDGCHB indicates that device matching or interface matching needs to be considered on a per-application basis. This PR switches the class to `IOUSBHostInterface` when the feature is enabled, allowing the interface number to be returned.

I am not familiar with macOS, but this seems like a fair compromise without changing the `serialport` APIs to make a distinction between searching for devices and searching for interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 ### Changed
 ### Fixed
+* Fix looking up `UsbPortInfo::interface` on macOS.
+  [#193](https://github.com/serialport/serialport-rs/pull/193)
 ### Removed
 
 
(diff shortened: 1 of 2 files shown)


sirhcel:
Thank you for spotting this @parasyte and for preparing this PR!

The `IOUSBHostDevice` class does not provide interface information but at a first glance does the `IOUSBHostInterface` class provide everything for `UsbPortInfo`. Since the feature gate `usbportinfo-interface` is scheduled to disappear with the next major release and the field `UsbPortInfo::interface`to become generally available.

After weighing this for while: Shouldn't looking up the information from `IOUSBHostInterface` instead from `IOUSBHostDevice` be the default?

parasyte:
So, again I need to be transparent that I don’t know Mac APIs. But my interpretation of the docs suggests that the Device and Interface classes should be used when you want either device or interface information. And that is going to be an application’s concern, not the library’s.

So, while it does appear, at least empirically, that the Interface class gives you everything you want… I’m not sure I would rely on that. They are different things for a reason. I suspect some devices might be capable of exposing information differently about itself (e.g. through the Device class) and about all of its interfaces (Interface class).

All in all, I’m in favor of just using the Interface class to simplify things. It can always be changed later if it breaks anything, right?

PR #195, opened 2024-06-06
Added support for 1.5 stop bits when reading current port settings

robdobsn:
On Windows 11 an Espressif ESP32-S3 module connected via USB defaults to 1.5 stop bits when in JTAG/CDC mode
Currently serialport-rs fails to open a port when it is in this state as the stop bits setting of 1.5 is not supported
This change does not allow 1.5 stop bits to be set but avoids failure to open when the port is already set to 1.5 stop bits

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,6 +225,9 @@ pub enum StopBits {
     /// One stop bit.
     One,
 
+    /// One and a half stop bits.
+    OnePointFive,
+
     /// Two stop bits.
     Two,
 }
@@ -233,6 +236,7 @@ impl fmt::Display for StopBits {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match *self {
             StopBits::One => write!(f, "One"),
+            StopBits::OnePointFive => write!(f, "OnePointFive"),
             StopBits::Two => write!(f, "Two"),
         }
     }
@@ -242,6 +246,7 @@ impl From<StopBits> for u8 {
     fn from(value: StopBits) -> Self {
         match value {
             StopBits::One => 1,
+            StopBits::OnePointFive => 1,
             StopBits::Two => 2,
         }
     }
--- a/examples/hardware_check.rs
+++ b/examples/hardware_check.rs
@@ -227,6 +227,7 @@ fn test_single_port(port: &mut dyn serialport::SerialPort, loopback: bool) {
     // Test setting stop bits
     println!("Testing stop bits...");
     stop_bits_check!(port, StopBits::Two);
+    stop_bits_check!(port, StopBits::OnePointFive);
     stop_bits_check!(port, StopBits::One);
 
     // Test bytes to read and write
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -326,6 +326,7 @@ impl SerialPort for COMPort {
         let dcb = dcb::get_dcb(self.handle)?;
         match dcb.StopBits {
             TWOSTOPBITS => Ok(StopBits::Two),
+            ONE5STOPBITS => Ok(StopBits::OnePointFive),
             ONESTOPBIT => Ok(StopBits::One),
             _ => Err(Error::new(
                 ErrorKind::Unknown,
--- a/src/windows/dcb.rs
+++ b/src/windows/dcb.rs
@@ -89,6 +89,7 @@ pub(crate) fn set_parity(dcb: &mut DCB, parity: Parity) {
 pub(crate) fn set_stop_bits(dcb: &mut DCB, stop_bits: StopBits) {
     dcb.StopBits = match stop_bits {
         StopBits::One => ONESTOPBIT,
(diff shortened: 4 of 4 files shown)


src/posix/termios.rs
```diff
@@ -198,6 +198,7 @@ pub(crate) fn set_data_bits(termios: &mut Termios, data_bits: DataBits) {
 pub(crate) fn set_stop_bits(termios: &mut Termios, stop_bits: StopBits) {
     match stop_bits {
         StopBits::One => termios.c_cflag &= !libc::CSTOPB,
+        StopBits::OnePointFive => termios.c_cflag &= !libc::CSTOPB,
```
Skgland, line 201: I don't think it is ideal to implicitly treat 1.5 as 1 here.
This might work for reading as we get a longer stop bit than we require, but for writing we would send a too short stop bit.

Based on [this stack overflow answer] in some circumstances treating 1.5 as 2 might even be expected. 

I don't think adding such a special case here is reasonably possible, so my suggestion would be to make this failable (i.e. return a `Result`) and return an `Err` for 1.5 on posix.

robdobsn: At the moment, if you specify that you want to open a serial port with 2 stop bits (or 1 stop bit for that matter) and the port is currently set to have 1.5 stop bits then the open fails. That makes the library unusable on devices that default to 1.5 stop bits. All I am trying to do is to stop this scenario from failing. I am not trying to add support for 1.5 stop bits and I don't think that would be possible (or desirable) without significant changes that probably go beyond this library.

Skgland: I'm not arguing against adding 1.5 stop bits. I just wanted to argue that treating 1.5 as 2 rather than 1 appears to be more consistent with other sources.
As such I would propose to either change the code to treat 1.5 as 2.0 or reject 1.5 (i.e. panic) for `set_stop_bits`/`<u8>::From<StopBits>::from` on unix. Either way I am not a maintainer of this project and the final decision lies in their not my purview.

robdobsn: I really don't care what happens when you try to set 1.5 stop bits. It could reject or treat as 1 or 2. That doesn't matter to me. What does matter is what happens when the port is already set to 1.5 stop bits. In this case currently you cannot do anything with the port. When you try to open the port (with 1 or 2 stop bits) it fails. That is what I am trying to get fixed.

src/lib.rs
```diff
@@ -242,6 +246,7 @@ impl From<StopBits> for u8 {
     fn from(value: StopBits) -> Self {
         match value {
             StopBits::One => 1,
+            StopBits::OnePointFive => 1,
```
Skgland, line 249:
```suggestion
            StopBits::OnePointFive |
```

src/posix/termios.rs
Skgland, line 201:
```suggestion
        StopBits::OnePointFive |
```

Skgland (commented):
This is effectively the change to this PR I am proposing.

sirhcel:
> Currently serialport-rs fails to open a port when it is in this state as the stop bits setting of 1.5 is not supported This change does not allow 1.5 stop bits to be set but avoids failure to open when the port is already set to 1.5 stop bits

Looking further into this issue, I can't find the path where opening a serial port (and starting to read some data) triggers interpreting the ports current stop bits configuration. Running [`examples/receive_data.rs`](https://github.com/sirhcel/serialport-rs/blob/5c0ede0b61496fa1987a96ae8b746213bb60570d/examples/receive_data.rs) with the interpretation of [any stop bit setting disabled](https://github.com/sirhcel/serialport-rs/commit/5c0ede0b61496fa1987a96ae8b746213bb60570d) gets to the point of waiting for incoming data:

```
>cargo run --example receive_data com8 115200
[...]
Receiving data on com8 at 115200 baud:
```

It looks like opening the port [always changes the ports stop bit setting](https://github.com/sirhcel/serialport-rs/blob/5c0ede0b61496fa1987a96ae8b746213bb60570d/src/windows/com.rs#L84) and this likely breaks communication with the  ESP32-S3. But I don't see where not supporting 1.5 stop bits makes opening the port fail (like in "with an error").

How are you attempting to open the port @robdobsn and which error are you getting in this case? Could you provide an example for reproducing this issue at best?


> As changing StopBits is a breaking change for the API, what about splitting this PR and tackling the issue in two steps:
> 
> 1. Reporting 1.5 stop bits as 2 when set to resolve the actual issue with the ESP32-S3

I'm asking because just [adding support for interpreting the 1.5 stop bit case](https://github.com/serialport/serialport-rs/blob/a9f8de8fe1f240d6520dbde0516092be926152b0/src/windows/dcb.rs#L92) would not make an issue with opening the port go away.

RossSmyth:
I did see this while reviewing other projects that does the opposite of moving to one stop bit
https://github.com/microsoft/AirSim/blob/6688d27d3712c2a9c824ababec7a2703475b6628/MavLinkCom/src/serial_com/SerialPort.cpp#L143-L145

But I think two makes more sense.

PR #197, opened 2024-06-18
fix #196

### LINKED ISSUE #196: Serial port operation has different performance in (windows10 \ MacOS)
The purpose of setting a timeout when using serialport scratch for serial communication is to prevent read operations from waiting indefinitely when no data arrives. If there is data, respond immediately and reset the timeout time. It is normal for me in MAC OS, but in Windows 10 system, even if there is data, I still need to wait until the timeout is reached before obtaining the data for the second read

```
use serialport;
use std::io::{self, Read, Write};
use std::thread;
use std::time::Duration;

use std::{
    // collections::HashMap,
    sync::{
        // mpsc::{Receiver, Sender, TryRecvError},
        Arc,
        Mutex,
    },
};

macro_rules! log {
    ($($arg:tt)*) => {
        println!(
            "{} ==> {}",
            chrono::Local::now().format("%Y-%m-%d %H:%M:%S%.3f"),
            format_args!($($arg)*)
        )
    };
}

fn main() {
    println!("Please press enter key：");

    let port_path = "COM1";
    let port = serialport::new(port_path, 115_200)
        .parity(serialport::Parity::None)
        .data_bits(serialport::DataBits::Eight)
        .stop_bits(serialport::StopBits::One)
        .flow_control(serialport::FlowControl::None)
        .timeout(Duration::from_millis(3000))
        .open()
        .unwrap();

    let shared_port = Arc::new(Mutex::new(port));

    let read_port = Arc::clone(&shared_port);

    let read_thread = thread::spawn(move || {
        let mut serial = read_port.lock().unwrap().try_clone().unwrap();
        loop {
            let mut buffer = [0u8; 2048];
            match serial.read(&mut buffer) {
                Ok(size) => {
                    log!("Read {} bytes: {:?}", size, &buffer[..size]);
                }

                Err(_err) => {
                    // log!("Error ! {:?}", &err);
                }
            };
            thread::sleep(Duration::from_millis(200));
        }
    });

    let write_port = Arc::clone(&shared_port);

    let write_thread = thread::spawn(move || {
        let value = [0x1b, 0x03, 0x00, 0x00, 0x1e];
        let mut serial = write_port.lock().unwrap().try_clone().unwrap();
        loop {
            let mut input = String::new();
            io::stdin().read_line(&mut input).unwrap();
            log!("<=== your press time {}", input);

            match serial.write(&value) {
                Ok(size) => {
                    log!("write success  size {}", size);
                }
                Err(error) => {
                    log!("write err  {:?}", &error);
                }
            };
            thread::sleep(Duration::from_millis(100));
        }
    });

    if let Err(e) = read_thread.join() {
        eprintln!("Read thread panicked: {:?}", e);
    }

    if let Err(e) = write_thread.join() {
        eprintln!("Write thread panicked: {:?}", e);
    }
}
```
windows
image
Mac
image

The same operation is very fast in Mac, and in Windows, it is necessary to reread and call the read method after timeout to obtain data, and the write operation is also very time-consuming.

antonio-quarta:
I think that this commit would fix the issue #196

As reported here:
https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-commtimeouts

> A value of MAXDWORD, combined with zero values for both the ReadTotalTimeoutConstant and ReadTotalTimeoutMultiplier members, specifies that the read operation is to return immediately with the bytes that have already been received, even if no bytes have been received.

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -10,7 +10,7 @@ use winapi::um::handleapi::*;
 use winapi::um::processthreadsapi::GetCurrentProcess;
 use winapi::um::winbase::*;
 use winapi::um::winnt::{
-    DUPLICATE_SAME_ACCESS, FILE_ATTRIBUTE_NORMAL, GENERIC_READ, GENERIC_WRITE, HANDLE,
+    DUPLICATE_SAME_ACCESS, FILE_ATTRIBUTE_NORMAL, GENERIC_READ, GENERIC_WRITE, HANDLE, MAXDWORD,
 };
 
 use crate::windows::dcb;
@@ -243,7 +243,7 @@ impl SerialPort for COMPort {
         let milliseconds = timeout.as_millis();
 
         let mut timeouts = COMMTIMEOUTS {
-            ReadIntervalTimeout: 0,
+            ReadIntervalTimeout: MAXDWORD,
             ReadTotalTimeoutMultiplier: 0,
             ReadTotalTimeoutConstant: milliseconds as DWORD,
             WriteTotalTimeoutMultiplier: 0,


sirhcel:
Thank you very much for looking into this and creating this PR @antonio-quarta! With #79, we've had a similar one around for quite a while and I've integrated this one about the time you created yours.

The changes from #79 already got released with [4.4.0](https://github.com/serialport/serialport-rs/releases/tag/v4.4.0). Does this solve the issue for you as well?

RossSmyth:
Yeah this has already been fixed in the latest release, so I would say this PR should probably be closed.

sirhcel:
> The changes from #79 already got released with [4.4.0](https://github.com/serialport/serialport-rs/releases/tag/v4.4.0). Does this solve the issue for you as well? Or is there a special reason for leaving `ReadTotalTimeoutMultiplier` at 0 instead of setting it to `MAXDWORD` as mentioned in the citation in the description?

I read the citation wrong in the first place. What it says is what we are doing with #79 and I will close #196.

antonio-quarta:
Sorry for late response, thank you for fixing this! 👍

PR #200, opened 2024-07-10
Add test for read returning available data before timeout

sirhcel:
This PR adds an integration test case checking that `read` returns the data currently available even before the specified read timeout has been reached.

This was not the case on Windows before #79 but there was no test covering this part of the expected behavior. According to the test added here, #196 should be fixed with release [4.4.0](https://github.com/serialport/serialport-rs/releases/tag/v4.4.0) and the behavior described in the issue can be seen when running this test on 0a75184.


eldruin (approved):
Great, thank you!

Issue #178, opened 2024-04-20
wrong serial number for FTDI modules ?


ffred:
Hi, I'm just beginning to use Serialport (a big thank you for your work ! 👍  ) and I notice a difference in the serial number reported, compare to the one reported by FT_Prog (from FTDI) or by USBDeview.

I tried it on Windows 10 & 11, with USB-RS485-PCB & USB-RS232-PCB modules. Serialport always give me the same serial number than FT-PROG / USBDeview, but with an added "A" at the end. for example FTFT5RXY12A instead of FT5RXY12.

not really a problem for me actually, I will use it with the extra "A", but just want to let you know.
thanks.
ffred

sirhcel:
Thank you for spotting this! I can confirm this behavior by running the example `list_ports` on both, macOS and Windows.

Interestingly, pyserial also reports an extra a at the end of the serial number on Windows, despite employing a dedicated parsing routine for [device identification strings starting with `FTDIBUS`](https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L373):
```
>python -m serial.tools.list_ports --verbose
[...]
COM6
    desc: USB Serial Port (COM6)
    hwid: USB VID:PID=0403:6001 SER=[...]A
2 ports found
```
The issue was already brought up with https://github.com/pyserial/pyserial/issues/61 and it ended up unresolved. The appended `A` seems to be produced by the driver stack and is already reported this way by `SetupDiGetDeviceInstanceIdA`. The Serial number reported by the USB device descriptor does not have the excess `A`.

A correct serial number seems to be reported by the "parent instance id". We could try to use this one but this would break serial number for users already looking for a certain one. A tough call ...

PR #170, opened 2024-03-28
Try find port type in parent device

soiamsoNG:
for docker enviroment  udev not run well for device add by --device 
without ID_BUS like ID* property, but we still can try find the must info from its parent.

below is sample output from udevadm info -t /dev/ttyACM0

> └─1-4:1.0
>           ┆ P: /devices/pci0000:00/0000:00:02.1/0000:01:00.0/usb1/1-4/1-4:1.0
>           ┆ E: DEVPATH=/devices/pci0000:00/0000:00:02.1/0000:01:00.0/usb1/1-4/1-4:1.0
>           ┆ E: SUBSYSTEM=usb
>           ┆ E: DEVTYPE=usb_interface
>           ┆ E: DRIVER=cdc_acm
>           ┆ E: PRODUCT=303a/1001/101
>           ┆ E: TYPE=239/2/1
>           ┆ E: INTERFACE=2/2/0
>           ┆ E: MODALIAS=usb:v303Ap1001d0101dcEFdsc02dp01ic02isc02ip00in00
>           └─tty/ttyACM0
>             ┆ P: /devices/pci0000:00/0000:00:02.1/0000:01:00.0/usb1/1-4/1-4:1.0/tty/ttyACM0
>             ┆ M: ttyACM0
>             ┆ R: 0
>             ┆ U: tty
>             ┆ D: c 166:0
>             ┆ N: ttyACM0
>             ┆ L: 0
>             ┆ E: DEVPATH=/devices/pci0000:00/0000:00:02.1/0000:01:00.0/usb1/1-4/1-4:1.0/tty/ttyACM0
>             ┆ E: SUBSYSTEM=tty
>             ┆ E: DEVNAME=/dev/ttyACM0
>             ┆ E: MAJOR=166
>             ┆ E: MINOR=0
>

PR #201, opened 2024-07-10
Remove Regex dependency

RossSmyth:
Regex being the slowest dependency for my entire project just to do a couple matches has been really bugging me. So I just recreated it.

The main draw back is the Unicode table. I just tried to recreate the semantics of the regex, and this is what it specified (any word character).

This is even lighter weight than regex-lite so I would say this succeeds #168, plus there is no MSRV junk to worry about (wrt dependencies). 

I don't personally believe there is any maintenance burden to it since the matching is pretty straightforward. 

I also added a test to test that we select the second part of serial numbers that are split by "&" as specified in one issue. That test seems to have been updated so that it no longer did that after parent IDs were added.

### DIFF
--- a/src/windows/enumerate.rs
+++ b/src/windows/enumerate.rs
@@ -68,7 +68,7 @@ fn get_ports_guids() -> Result<Vec<GUID>> {
     Ok(guids)
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 struct HwidMatches<'hwid> {
     vid: &'hwid str,
     pid: &'hwid str,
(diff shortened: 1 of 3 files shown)


parasyte:
💯

I have a pet peeve with static RegExs. They add a lot of overhead compared to a simple FSM.

sirhcel:
Thank you for keeping on working on this issue @RossSmyth! The call for regex support is also imminent with #170 but in looks like we can get away there without regex or regex-lite there as well.

After a quick look at HWIDs and the MODALIAS strings from #170, I'm wondering if we actually need to deal with word boundaries according to Perl regex or if we could use a simpler definition in our context? I'm not opposed to a solution based on `PERL_WORD` but with this big step already made, isn't there a even more compact solution within reach?

RossSmyth:
> I'm wondering if we actually need to deal with word boundaries according to Perl regex or if we could use a simpler definition in our context?

Yeah I was thinking the same, but I personally do not know what that subset is. The Windows documentation on HWID isn't really that great.

RossSmyth:
Okie dokie. I'll modify it so.

src/windows/enumerate.rs
```diff
@@ -504,7 +549,7 @@ fn test_parsing_usb_port_information() {
 
     assert_eq!(info.vid, 0x0403);
     assert_eq!(info.pid, 0x6001);
-    assert_eq!(info.serial_number, Some("A702TB52A".to_string()));
+    assert_eq!(info.serial_number, Some(r"A702TB52A\0000".to_string()));
```
RossSmyth, line 552: With the new impl, if I interpreted what you & the pyserial regex mean correctly, it just matches to the end of the string. Meaning that this is the serial number it interprets now.

sirhcel: Wow, you are quick! Yes, pySerial's regex matches until the end of the string. But they are distinguishing between the two bus types `USB` and `FTDIBUS` for the former we are sorted. But for the latter we need to stop matching at the end of the middle paths segment and ignore `\0000`.

sirhcel: May be splitting the string at first at `\`, then deciding how to interpret middle path element according to the first one (`FTDIBUS` and everything else) and then handle things according to whether we are dealing with the interface of a composite device or the device itself.

RossSmyth:
Thinking about how the serial number is matched, we currently make the serial number an optional field, even from the parent device if part of a composite device. We do not test this behavior though. So I wonder what the semantics are supposed to be. 

1. Serial numbers are always optional (this is what is currently implemented)
2. Serial numbers of child devices of composite devices are optional, but the parent will always have a serial
3. Something weirder?

RossSmyth:
Ok, my Python skills are quite rusty but I think it's starting right in the face in your link.
https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L337-L343

I believe it is supposed to be that if the child device doesn't have a serial number, the parent must, or else pyserial will throw an exception OR stack overflow because it seems to just infinitely recurse 
https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L230-L231

sirhcel:
> Thinking about how the serial number is matched, we currently make the serial number an optional field, even from the parent device if part of a composite device. We do not test this behavior though. So I wonder what the semantics are supposed to be.
> 
>     1. Serial numbers are always optional (this is what is currently implemented)

As far as I've overseeing things here: This is is the way. Serial numbers for USB devices are provided by the device and are optional. In terms of USB descriptors the serial number is a property of the USB device and not the interface. In case of a composite device there is one serial number per device and the interfaces are numbered.

This is also related to #178 and we are maybe using the wrong serial number for composite FTDI devices: The suffix seems to be generated by the FTDI driver and according to https://github.com/pyserial/pyserial/issues/283#issue-274725473 (COM96 and COM134), the parent device provides the right serial number.

RossSmyth:
> the parent device provides the right serial number.

If you look at `get_parent_serial_number()` that does not seem to be the guaranteed case.
https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L162-L164

So yeah, I agree. Always optional.

sirhcel:
> I believe it is supposed to be that if the child device doesn't have a serial number, the parent must, or else pyserial will throw an exception OR stack overflow because it seems to just infinitely recurse https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L230-L231

Recursion depth is limited [here](https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L163).


> I'll change this in a later PR though as it is a bit orthogonal.

Yes. No need to change/fix everything here. We just should end up at least en par with the beforehand state.

sirhcel:
> If you look at `get_parent_serial_number()` that does not seem to be the guaranteed case. https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_windows.py#L162-L164

Like in: No serial if the actual device is nested too deep down in the device stack?

> So yeah, I agree. Always optional.

Great to be on the same page! 😄

sirhcel:
I would like to keep the actual wrong serial numbers returned in #203 and not return a longer number which is also not the devices serial number.

Switching from `char::is_ascii_alphanumeric` to `char::is_alphanumeric` should bring us a bit more toward `PERL_WORD` with still just code from the standard library. Do you see an issue with that? Did you use `is_ascii_alphanumeric` for a certain reason?

Locally testing the changes shows no difference in the output for my set of test hardware. From my perspective, we are good to go. Thank you very much for your perseverance with this PR!

RossSmyth:
I used it because it is semantically identical to the pyserial regex. But we can preserve the PERL_WORD semantics instead, that works for me.

RossSmyth:
Oops forgot to actually add the changes 😄 

It did not handling actual Unicode in the serial number properly if we are to accept that. The previous implementation would index it by byte. For for example the test serial I added, `3854356β03432`, would be matched as `3854356β0343` since it would index it by byte and have the end chopped off. That is fixed with tests that poke it with the edge cases I found while implementing.

sirhcel:
Thank you for finally remembering and adding the last fix!

RossSmyth:
Thanks!

PR #202, opened 2024-07-17
fix(usb): improve robustness of interface field when using IOKit backend

doinkythederp:
This PR improves the robustness of the code that finds the USB port interface number. Instead of accessing the nonexistent `IOUSBHostDevice.bInterfaceNumber` property, it accesses `IOUSBHostInterface.bInterfaceNumber`.


doinkythederp:
It actually looks like this has the same behavior as #193.

sirhcel:
Thank you for spotting this issue and preparing a PR @doinkythederp!

> It actually looks like this has the same behavior as #193.

I already integrated #193 and this should give the same results. Could you please give https://github.com/serialport/serialport-rs/commit/503905074ffea98811aec9974d805a1f65d07319 a spin and reopen this issue if things aren't solved for you?

PR #208, opened 2024-07-31
Clamp long timeouts (like Duration::MAX)

### LINKED ISSUE #207: SerialPortBuilder.timeout(Duration::MAX) gives "Invalid argument"
When serial port is created through `SerialPortBuilder` and `.timeout()` is called with the argument set to `Duration::MAX`, any attempt to read from the port returns an error `Custom { kind: Other, error: "Invalid argument" }`.
Apparently there is no other way to set blocking read without a timeout.

Code to reproduce:
```rust
use std::{time::Duration, error::Error};

fn main() -> Result<(), Box<dyn Error>> {
    let mut serial = serialport::new("/dev/ttyUSB0", 9600)
        .timeout(Duration::MAX)
        .open()?;
    let mut tmp = vec![0];
    serial.read(&mut tmp)?;
    dbg!(tmp);
    Ok(())
}
```

Environment:
* Fedora 40
* 6.9.11-200.fc40.x86_64
* rustc 1.79.0 (129f3b996 2024-06-10)
* serialport = "4.4.0"
* `/dev/ttyUSB0`: ID 1a86:7523 QinHeng Electronics CH340 serial converter 
* also reproducible with `/dev/tty`

sirhcel:
This PR fixes #207 by clamping long timeouts to the capabilities of the underlying platform. Thin preserves monoticity for for the conversion of `Duration` values and does not expose the details of the conversion methods previously used for the different platforms like:

* Panics from overflows
* Non-monotonic actual timeouts duo to cutting off higher bits/modulo operations from casts
* Unexpected errors when opening ports or reading data 

This is especially unpleasant as we are currently recommending using `Duration::MAX` as "no timeout" (#12).

In theory, this trades accuracy of long timeouts in. In practice, long timeouts already exhibited unexpected and faulty behavior from returning error to panicking due to arithmetic overflows.

It looks like no platform uses an actual timeout data type smaller than `i32`. So even in combination with expressing timeouts as milliseconds, this gives about 24 days for which timeouts could be expressed accurately. Everything beyond can likely be handled witch lesser precision and spurious timeouts.


eldruin (approved):
Very thorough. Thank you!

Issue #156, opened 2024-02-03
Set a Custom Baudrate


thekief:
I implemented a small application to interact with a serial port that controls some stuff. While the involved USB-to-UART converter (CH340N)  officially seems to support only the common baud rates, the original driver for this device used a baud rate of `10000`. For whatever reason that seems the only baud rate to be reliably working. 

While working on my application I noticed that the **serialport** library seems to default to a baud rate of `9600`, atleast that is what `port.baud_rate().unwrap()` returns. I dumped the systemcalls of a reliably working Python script and my application and noticed following call missing from my application:

```
ioctl(3, TCGETS, {c_iflag=IGNPAR, ... ) = 0
ioctl(3, TCGETS2, ..., c_ispeed=10000, c_ospeed=10000}) = 0
ioctl(3, TCSETS2, ... , c_ispeed=10000, c_ospeed=10000}) = 0
```

Neither setting the baud rate with `serialport::new` nor setting the value before opening it with `baud_rate` seems to change anything. Any suggestions how I could get that custom baud ate properly running?

mrcbax:
Here is some good reference C code that can do this:
https://github.com/Robokishan/custom-baud

I need serialport-rs to support setting arbitrary baud rates.

PR #213, opened 2024-08-30
Don't ignore errors from setting baud rates

sirhcel:
A [chat](https://matrix.to/#/!cnJwoUGZmLziNcGNua:matrix.org/$YfHJlCTo6T0WLg85wWWhF4-Niiq7b8gn9njXYD5gMXg?via=matrix.org&via=t2bot.io&via=matrix.shymega.org.uk) a while ago led to the discovery that we are currently [silently ignoring](https://github.com/serialport/serialport-rs/blob/18f5c2ab5e2b08d839fd6dd0d4c5624778da593b/src/posix/termios.rs#L284) attempts to set an unsupported baud rate on Linux/musl.

This PR changes the behavior to report an error in this case.

Even though this does not add the support, it will no longer keep users in the dark about this issue like it likely happened in #156.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 ### Changed
 ### Fixed
+* Fix ignoring errors from setting baud rate and ignoring unsupported baud
+  rates.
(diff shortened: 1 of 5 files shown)


eldruin (approved):
Looks good to me, thank you!

PR #216, opened 2024-09-05
Cleanup leftover dbg from clamping timeouts

sirhcel:
We should not generate any output to stdout or stderr unconditionally. Especially as a library.

This PR brings just the cleanup for #215 to buy some time for the discussion there.


eldruin (approved):
Looks good to me, thank you!

Issue #214, opened 2024-08-31
macos and Windows disagree on interface numbering for Black Magic Probe


xobs:
The Black Magic Probe has two serial ports, and the interface should be 0 and 2. For example, this is how it shows up under Windows:

```
   Compiling serialport v4.5.1-alpha.0 (E:\Code\serialport-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.71s
     Running `target\debug\examples\list_ports.exe`
Found 2 ports:
  COM8
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Microsoft
           Product: USB Serial Device (COM8)
         Interface: 02
  COM9
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Microsoft
           Product: USB Serial Device (COM9)
         Interface: 00
```

However under Mac, it lists each port twice, and the interface is offset by one:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/list_ports`
Found 4 ports:
  /dev/cu.usbmodem97B6DC141
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Black Magic Debug
           Product: Black Magic Probe v1.10.0-634-gfa79ab6b
         Interface: 01
  /dev/tty.usbmodem97B6DC141
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Black Magic Debug
           Product: Black Magic Probe v1.10.0-634-gfa79ab6b
         Interface: 01
  /dev/cu.usbmodem97B6DC143
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Black Magic Debug
           Product: Black Magic Probe v1.10.0-634-gfa79ab6b
         Interface: 03
  /dev/tty.usbmodem97B6DC143
    Type: USB
    VID:1d50 PID:6018
     Serial Number: 97B6DC14
      Manufacturer: Black Magic Debug
           Product: Black Magic Probe v1.10.0-634-gfa79ab6b
         Interface: 03
```

There are two issues:

1. Serial ports are doubled, with both the "cu" and "tty" variant presented, and
2. The Interface is offset by one

[...]

PR #219, opened 2024-09-26
usbportinfo: document interface variation

xobs:
Document that the interface number can vary depending on the platform. This addresses some surprise discovered in #214, as it isn't entirely clear which interface should be used, but it is definitely clear that different platforms report differing interface indexes.


sirhcel:
Thank you very much for your swift contribution!

PR #220, opened 2024-10-05
linux: support getting usb port info without libudev

xobs:
libudev can make cross-compiling difficult, particularly due to its reliance on pkg-config. It is used primarily for enumeration.

Support getting USB port info such as VID and PID when libudev is disabled. This reads this information out of /sys/. If the information cannot be discovered, then the port type will be `Unknown`.

The method of discovery is based on pySerial, which does something similar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ project adheres to [Semantic Versioning](https://semver.org/).
 
 * Add recommendation on how to interpret `UsbPortInfo::interface_number`.
   [#219](https://github.com/serialport/serialport-rs/pull/219)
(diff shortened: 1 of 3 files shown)


sirhcel:
This is a great improvement. Looks good (at initial testing). Thank you very much for your contribution!

SergioGasquez:
Thanks for this great feature! It might be worth to update the readme as it says: 
> The implementation on Linux using glibc relies on libudev, an external dynamic library that will need to be available on the system the final binary is running on. 

Which is only true when using the `libudev` feature.

eldruin:
Good point @SergioGasquez, could you submit a PR explaining that?

Issue #165, opened 2024-03-11
Tracking issue for WebSerial support.


shymega:
I'm working with the PineBuds, and in discussions with a team member, we deliberated whenever we could use this crate with the [WebSerial](https://developer.mozilla.org/en-US/docs/Web/API/Web_Serial_API) API.

I was thinking of adding a trait impl, guarded by the `wasm` Rust target so that we could use it that way.

I'm just drafting this issue tonight, but I'm a little sleepy, so I'll re-address it when I'm functional again.

Thanks!

RossSmyth:
Note that it is not portable and Mozilla has indicated they never plan on supporting it.
mozilla/standards-positions#336

robo9k:
> Note that it is not portable and Mozilla has indicated they don't plan on supporting it. [mozilla/standards-positions#336](https://github.com/mozilla/standards-positions/issues/336)

That position has changed and Web Serial is available in Firefox Desktop, starting with version 151
https://hacks.mozilla.org/2026/05/web-serial-support-in-firefox/

PR #224, opened 2024-10-23
[WIP]: Implement WebSerial/wasm

i404788:
Resolves https://github.com/serialport/serialport-rs/issues/165. As stated in that issue it's not supported on all browsers although it has been 'stable' on those browsers since 2021.

Since WebSerial is entirely Promise based there is a lot of `block_on` to make it sync.

Currently status:
* [x] Core serial operations and options
* [ ] Timeouts
* [ ] Flow-control Inputs
* [ ] `*_to_read` (see inline comment)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,6 @@
 # by default to avoid race conditions (see
 # https://github.com/rust-lang/cargo/issues/8430).
 RUST_TEST_THREADS = "1"
+
+[target.wasm32-unknown-unknown]
+rustflags = ["--cfg=web_sys_unstable_apis"]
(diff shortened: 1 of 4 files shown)


sirhcel:
Thank you opening this draft PR @i404788! I know WebSerial only from hearsay. When progressing this PR, could you add an example for how to use WebSerial support?

i404788:
Hey @sirhcel,

Unfortunately we've hit a bit of a snag with the `block_on`, since it doesn't yield back to JS it will never complete the promises. It seems to be impossible at the moment to finish this PR 

In theory if you could interrupt the wasm task a specific impl of block_on could exist, although I'm not familiar enough with wasm internals to know the feasibility of that. e.g. you could checkpoint with https://jubianchi.gitbook.io/wasmer-docs/glibc-reqs/integrations/examples/exit-early, there was a discussion about this but ended up not being spec'ed: https://github.com/WebAssembly/design/issues/712

For now I'll close the PR as I don't believe this will be added in a generic way any time soon. (this also means https://github.com/serialport/serialport-rs/issues/165 is impossible atm given the goals of this project).

Issue #177, opened 2024-04-11
Potentially confusing read_exact timeout semantics


RossSmyth:
Since this library uses the [`default_read_exact`](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/std/io/mod.rs.html#465-482) implementation, `read_exact` has different timeout semantics compared to `read`. Since the implementation of this default method is to call `read` in a loop, each loop has a new timeout. This could potentially lead to confusing timeouts:

1. Set some baud & timeout. Say 9600 bps and 110ms
2. Call read_exact with 100 bytes. This would mean it would take about 100ms for a full transfer if each byte is contiguous
3. In the degenerate case, one byte could come every 109ms, leading to the method call hanging for ~11s.

At minimum this should be documented somewhere.

sirhcel:
Thank you very much for spotting this issue! Documenting would be good and providing an as documentation is seldomly read, an implementation of `read_exact` following the principle of least surprise would be even better.

PR #225, opened 2024-10-23
handle EINTR return code from tcdrain to effectively flush

ndusart:
Hi, this PR adds the handling of EINTR return code from `tcdrain`.

If returning EINTR, it does not mean an actual error for the flushing operation and this should be retried.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ project adheres to [Semantic Versioning](https://semver.org/).
   [#238](https://github.com/serialport/serialport-rs/pull/238)
 
 ### Fixed
+
+* Retry flushing data on `EINTR` up to the ports read/write timeout.
+  [#225](https://github.com/serialport/serialport-rs/pull/225)
+
 ### Removed
 
 
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -441,8 +441,13 @@ impl io::Write for TTYPort {
     }
 
     fn flush(&mut self) -> io::Result<()> {
-        nix::sys::termios::tcdrain(self.fd)
-            .map_err(|_| io::Error::new(io::ErrorKind::Other, "flush failed"))
+        loop {
+            return match nix::sys::termios::tcdrain(self.fd) {
+                Ok(_) => Ok(()),
+                Err(nix::errno::Errno::EINTR) => continue,
+                Err(_) => Err(io::Error::new(io::ErrorKind::Other, "flush failed")),
+            };
+        }
     }
 }
 
--- a/src/posix/tty.rs
+++ b/src/posix/tty.rs
@@ -1,7 +1,7 @@
 use std::mem::MaybeUninit;
 use std::os::unix::prelude::*;
 use std::path::Path;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use std::{io, mem};
 
 use nix::fcntl::{fcntl, OFlag};
@@ -441,10 +441,22 @@ impl io::Write for TTYPort {
     }
 
     fn flush(&mut self) -> io::Result<()> {
+        let timeout = Instant::now() + self.timeout;
         loop {
             return match nix::sys::termios::tcdrain(self.fd) {
                 Ok(_) => Ok(()),
-                Err(nix::errno::Errno::EINTR) => continue,
+                Err(nix::errno::Errno::EINTR) => {
+                    // Retry flushing. But only up to the ports timeout for not retrying
+                    // indefinitely in case that it gets interrupted again.
+                    if Instant::now() < timeout {
+                        continue;
+                    } else {
+                        Err(io::Error::new(
+                            io::ErrorKind::TimedOut,
+                            "timeout for retrying flush reached",
+                        ))
+                    }
+                }
                 Err(_) => Err(io::Error::new(io::ErrorKind::Other, "flush failed")),
             };
         }


sirhcel:
Thank you for looking into this and opening this PR!

Recently there was a discussion regarding `EINTR` and returning errors in https://github.com/serialport/serialport-rs/pull/215#issuecomment-2363642307. We discussed this in the context of [`std::io::Read::read`](https://doc.rust-lang.org/1.82.0/std/io/trait.Read.html#errors) and for this operation it was clearly stated that `read` is supposed to return an error with `ErrorKind::Interrupted` in this case.

The same is stated for [`std::io::Write::write`](https://doc.rust-lang.org/1.82.0/std/io/trait.Write.html#errors) and from my gut instinct I would have expected it to be the same for `flush`. But [it isn't](https://doc.rust-lang.org/1.82.0/std/io/trait.Write.html#tymethod.flush). So repeating the underlying flush operation is likely the right thing to do in this case like for example the default implementation of `Write::write_all` [documents](https://doc.rust-lang.org/1.82.0/std/io/trait.Write.html#method.write_all) and [does](https://doc.rust-lang.org/1.82.0/src/std/io/mod.rs.html#1687-1699). Let me sleep over it.

Nonetheless, the (previously) returned error kind [`ErrorKind::Other`](https://github.com/serialport/serialport-rs/pull/225/files#diff-461484591c9660e1a50a6140bdf1dac15e5f3124dd5335e82d93e2e75c08709aL445) is not what I would have expected here nor does the documentation for `flush` mandate such a behavior. Shouldn't the underlying error from `tcdrain` being "bubbled up" like in is [done for `read`](https://github.com/serialport/serialport-rs/blob/f854b8d5bf8593e33461f50ab666996f4bc3a1d4/src/posix/tty.rs#L437) with [`From<nix::Error> for Error`](https://github.com/serialport/serialport-rs/blob/f854b8d5bf8593e33461f50ab666996f4bc3a1d4/src/posix/error.rs#L18)?

ndusart:
@sirhcel if the specific error is forwarded as an `ErrorKind::Interrupted` that would indeed be a solution too, the main issue here is that this error cannot be differentiated in regards to other errors.

It is probably a better approach to indeed let the client side decide how to handle this error. 

Maybe we should instead have a method in `std::io::Write` trait similar to `write_all` but for `flush`. But, definitely not the place to discuss this ^^'

I'd favor your proposition then :)

sirhcel:
> @sirhcel if the specific error is forwarded as an `ErrorKind::Interrupted` that would indeed be a solution too, the main issue here is that this error cannot be differentiated in regards to other errors.

Then we agree on the point that flush should return distinguishable error values. Good.


> It is probably a better approach to indeed let the client side decide how to handle this error.

Under which circumstances are you getting `EINTR` in `flush`?

I asked the question on how `flush` is expected to behave in the [t-libs Zulip channel](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/.E2.9C.94.20ErrorKind.3A.3AInterrupted.20from.20std.3A.3Aio.3A.3AWrite.3A.3Aflush) and the response made a point for retrying.

In case of _Ctrl_ + _C_ automatically retrying at "library level" might interfere with the user's intention of stopping the program execution (if flushing takes a significant amount of time). Returning `ErrorKind::Interrupted` lays this decision into the hands of the application. There have been no other issues with errors from `flush` so far and so I'm assuming that returning `ErrorKind::Interrupted` in this case is fine.


> Maybe we should instead have a method in `std::io::Write` trait similar to `write_all` but for `flush`. But, definitely not the place to discuss this ^^'
> 
> I'd favor your proposition then :)

Then let's "bubble up" the right error, document this behavior and we should be sorted here.

ndusart:
> Under which circumstances are you getting `EINTR` in `flush`?

I have a library using `serialport` exposing a blocking API to communicate with a specific device and we export a C API to reuse it on multiple projects. We have this behaviour when the library is used in a Flutter application.

Flutter application runs on a asynchronous event loop and I guess application must receive some signals that are sent internally by the flutter engine to drive the loop.

You make a point about signal from Ctrl+C. And basically looping through `tcdrain` would make the same issue about timeouts as it is for `read_exact` or `write_all` (#177)

> Then let's "bubble up" the right error, document this behavior and we should be sorted here.

Yes indeed. Do you prefer that I rebase the PR with this fix or will you do it yourself and we close this ?

sirhcel:
Thank you for the information on the background of this PR!

> You make a point about signal from Ctrl+C. And basically looping through `tcdrain` would make a similar issue as the timeout issue for `read_exact` or `write_all` (#177) and potentially could deadlock in some situations.

Sleeping over it for some nights put my decisiveness a step back. Because - as you're mentioning here - `read_exact` and `write_all` are retrying on `ErrorKind::Interrupted`. So we are already bought into retrying in with these methods.

Let's assume for a moment that #177 is fixed.  The execution time of `write_all` and `read_exact` would be limited by this timeout in the case of `ErrorKind::Interrupted` from the underlying `write`/`read`. So what about applying the same semantics to `flush` as well? Retrying, but with timeout as the limit? Wouldn't this behavior follow the documentation of `Write::flush` more closely than returning `ErrorKind::Interrupted` for `EINTR`?

Timeout behavior is not specified by `Read` and `Write`. But timeouts are a specialty of our implementation for serial ports and could be seem as something applying to all `Read` and `Write` operations.

What's your opinion on this @ndusart?

> Yes indeed. Do you prefer that I rebase the PR with this fix or will you do it yourself and we close this ?

Would be great if you could lend a hand.

Issue #222, opened 2024-10-18
Physical port for usb serial devices


anti-social:
I'm working on a program that flashes and sets up an stm32 mcu device. During a setup the device should be rebooted multiple times. So I need a way to identify the same device between reboots. USB physical port is the only thing that is stable. For example `nusb` crate has `bus_id` and `port_chain` for physical port identification (they don't release it yet): [port_chain](https://github.com/kevinmehall/nusb/commit/2c9c699a286b6b23317a9f026ec06279b2bb8dac#diff-1bbf4db47857068ae156bef3d4ad717c9e9888956e064c16456c23b1491e5c84R126)

It would be nice to have similar data in a `UsbPortInfo` struct.

RichardWGNR:
> I'm working on a program that flashes and sets up an stm32 mcu device. During a setup the device should be rebooted multiple times. So I need a way to identify the same device between reboots. USB physical port is the only thing that is stable. For example `nusb` crate has `bus_id` and `port_chain` for physical port identification (they don't release it yet): [port_chain](https://github.com/kevinmehall/nusb/commit/2c9c699a286b6b23317a9f026ec06279b2bb8dac#diff-1bbf4db47857068ae156bef3d4ad717c9e9888956e064c16456c23b1491e5c84R126)
> 
> It would be nice to have similar data in a `UsbPortInfo` struct.

How did you solve this problem?

xobs:
I'd like to see this feature implemented. Do you have any interest in continuing it? If not, would you mind if I took a crack at it?

One issue I see with your patch is that it is not semver-compatible and would require a new minor version. Still, I think it's a good way to implement it.

xobs:
I put together https://github.com/serialport/serialport-rs/pull/350 which takes a slightly different approach on Windows but otherwise is similar.

PR #227, opened 2024-11-24
Add bus_id and port_chain field to UsbPortInfo (all platforms)

kindermax:
Add `bus_id` and `port_chain` field to UsbPortInfo (all platforms)

Implementation taken and adapted from `nusb` crate

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -787,6 +787,8 @@ pub struct UsbPortInfo {
     pub manufacturer: Option<String>,
     /// Product name (arbitrary string)
     pub product: Option<String>,
+    /// Location (bus_id-a.b string)
+    pub location: String,
     /// The interface index of the USB serial port. This can be either the interface number of
     /// the communication interface (as is the case on Windows and Linux) or the data
     /// interface (as is the case on macOS), so you should recognize both interface numbers.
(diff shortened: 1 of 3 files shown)


sirhcel:
Thank you for starting this PR! We've currently got a (CI) issue with one of our sub dependency which is about getting resolved with #229 and #230. So rebasing the changes on #230 will resolve the red herrings arising from #229 in the CI builds here as well.

Have you completely abandoned this idea or are you still working on it in https://github.com/kindermax/serialport-rs/pull/1? In case you are, do you know if there is a cross-platform string representation of a port "path"?

sirhcel:
Have you seen the reference to the nusb crate regarding port chains #222? They seem to use a binary representation. Do you see a big plus for either a string representation or a vector of port numbers?

I have not looked into it but using nusb for determining bus ID and port chain might also be an option.

kindermax:
Hi, @sirhcel, I am still working on this functionality, I just mistakenly opened PR to the original repo instead of my fork, because the work wasn't ready to be shown here yet. But, anyway)

In general, I tried to implement this functionality as close as possible to `nusb` to be able to share the same location representation between `nusb` part of a program and `serialport-rs` part of a program.

kindermax:
> I have not looked into it but using nusb for determining bus ID and port chain might also be an option.
Theoretically this is possible, if somehow match `nusb::DeviceInfo` with `SerialPortType::UsbPort(UsbPortInfo)` then we indeed can use port info from `nusb`, but I went in other direction and end up with this PR)

kindermax:
Reopened PR

PR #229, opened 2024-12-01
Pin libc subdependency for MSRV 1.59.0 and allow successor to Unicode DFS license

sirhcel:
* The libc crate bumped MSRV from 1.19.0 to 1.63.0 with patch release 0.2.164
    * This breaks our builds (like run [12104758738](https://github.com/serialport/serialport-rs/actions/runs/12104758738)) as we are still using MSRV 1.59.0
    * Adding a direct dependency to massage the dependency resolver into using a compatible version already did the trick with  #186
* In addition the license of the sub dependency unicode-ident got updated
    * We allowed the Unicode DFS license
    * Unicode 3.0 is a successor to Unicode DFS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 ### Changed
 ### Fixed
+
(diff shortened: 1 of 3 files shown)


eldruin (approved):
Thank you!

PR #233, opened 2024-12-12
impl `Sync` for `SerialPort` trait

### LINKED ISSUE #232: Consider `impl Sync` for `SerialPort` trait
I would like to propose implementing the `Sync` trait for the `SerialPort` trait. The primary blocker for this seems like `COMPort` does not automatically implement `Sync` due to the `HANDLE` type using a raw pointer. `TTYPort` already implements `Sync` because it uses `RawFd` which is simply a primitive type.

Upon investigation, I found the [Rust standard library](https://doc.rust-lang.org/src/std/os/windows/io/handle.rs.html#111-117) contains the following statement about Windows `HANDLE` types:

```rust
// The Windows [`HANDLE`] type may be transferred across and shared between
// thread boundaries (despite containing a `*mut void`, which in general isn't
// `Send` or `Sync`).
//
// [`HANDLE`]: std::os::windows::raw::HANDLE
#[stable(feature = "io_safety", since = "1.63.0")]
unsafe impl Send for OwnedHandle {}
// ...
#[stable(feature = "io_safety", since = "1.63.0")]
unsafe impl Sync for OwnedHandle {}
```

Maybe we could do the same for the `COMPort` type by implementing the `Sync` trait similarly. This would allow `SerialPort` to also implement `Sync`, enabling usage in multi-threaded contexts. I believe this is sound because IO-related operations still require exclusive access to `&mut Self` to perform.

unlimitedsola:
Closes #232 

There are many alternatives that don't involve `unsafe impl`, but I believe this is the simplest one that doesn't involve an MSRV bump, nor should it introduce breakage, at least I hope...

One alternative is to use [`OwnedHandle`](https://doc.rust-lang.org/std/os/windows/io/struct.OwnedHandle.html) provided by stdlib. However, doing so requires an MSRV bump to at least 1.63.0, which I've heard would require a major version bump in this library so I refrained from this approach.

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -32,11 +32,9 @@ pub struct COMPort {
     port_name: Option<String>,
 }
 
-// The Windows [`HANDLE`] type is considered safe according to the standard library.
-// See explanation on [`OwnedHandle`] in stdlib for more information.
-//
-// [`HANDLE`]: std::os::windows::raw::HANDLE
-// [`OwnedHandle`]: std::os::windows::io::OwnedHandle
+// The Windows `HANDLE` type is considered safe according to the standard library.
+// See the explanation below in stdlib for more information:
+// https://github.com/rust-lang/rust/blob/f4f0fafd0c7849e162eddbc69fa5fe82dbec28c7/library/std/src/os/windows/io/handle.rs#L111-L115
 //
 // In the future we might want to consider using `OwnedHandle` instead of `HANDLE` directly.
 // At the time of writing, such work is blocked by this crate's MSRV (1.59.0) policy as
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -32,7 +32,17 @@ pub struct COMPort {
     port_name: Option<String>,
 }
 
+// The Windows [`HANDLE`] type is considered safe according to the standard library.
+// See explanation on [`OwnedHandle`] in stdlib for more information.
+//
+// [`HANDLE`]: std::os::windows::raw::HANDLE
+// [`OwnedHandle`]: std::os::windows::io::OwnedHandle
+//
+// In the future we might want to consider using `OwnedHandle` instead of `HANDLE` directly.
+// At the time of writing, such work is blocked by this crate's MSRV (1.59.0) policy as
+// `OwnedHandle` was stabilized in 1.63.0.
 unsafe impl Send for COMPort {}
+unsafe impl Sync for COMPort {}
 
 impl COMPort {
     /// Opens a COM port as a serial device.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -447,7 +447,7 @@ impl SerialPortBuilder {
 ///
 /// This trait is all that's necessary to implement a new serial port driver
 /// for a new platform.
-pub trait SerialPort: Send + io::Read + io::Write {
+pub trait SerialPort: Send + Sync + io::Read + io::Write {
     // Port settings getters
 
     /// Returns the name of this port if it exists.


src/windows/com.rs
```diff
@@ -32,7 +32,17 @@ pub struct COMPort {
     port_name: Option<String>,
 }
 
+// The Windows [`HANDLE`] type is considered safe according to the standard library.
+// See explanation on [`OwnedHandle`] in stdlib for more information.
```
sirhcel, line 36: Looking at https://doc.rust-lang.org/std/os/windows/raw/type.HANDLE.html or https://doc.rust-lang.org/std/os/windows/io/struct.OwnedHandle.html, I don't see the documentation you are mentioning at al first glance. Where is the statement made you are referring to? In [this comment in `handle.rs`](https://doc.rust-lang.org/src/std/os/windows/io/handle.rs.html#111)

sirhcel (commented):
Thank you very much for this PR! Please see my comment below. I'm wondering what's you use case for sharing `&SerialPort` between threads? At a firs glance there are only a few meaningful things to do with an immutable reference.

src/windows/com.rs
unlimitedsola: Yes, I'm referring to the comment in [`handle.rs`](https://doc.rust-lang.org/src/std/os/windows/io/handle.rs.html#111-117) that you linked.
I'm not sure how to link it to an `impl` block so I figured the definition of `OwnedHandle` might be the closest to that comment (50+ lines away)

The link to the std source https://doc.rust-lang.org/src/std/os/windows/io/handle.rs.html#111-117 doesn't seem to be a permanent one so I was worried putting the doc link in might become invalid in the future.

I can probably replace it with a GitHub permalink like this: https://github.com/rust-lang/rust/blob/f4f0fafd0c7849e162eddbc69fa5fe82dbec28c7/library/std/src/os/windows/io/handle.rs#L111-L115
but that would be a very long line unfortunately.

unlimitedsola:
> I'm wondering what's you use case for sharing `&SerialPort` between threads? At a firs glance there are only a few meaningful things to do with an immutable reference.

I was trying to display some port information and settings that should only require an immutable reference in another thread. Without `impl Sync`, however, it will force me to pass _ownership_ to that thread and then get it back.

src/windows/com.rs
sirhcel: Alight. A permalink would be great then because a long line looks way more appealing to me than the risk of getting it broken at some point in time.

unlimitedsola: Thanks for your feedback! I've now updated the patch with a permalink to the explanation.

okhsunrog:
Is there anything that blocks this from getting merged?

okhsunrog:
@sirhcel I used this crate via tokio-serial crate, and it worked fine on Linux. Once I tried to compile my app for Windows, I saw the error that SeriaPort is not Sync on windows. So, for now I have this in my code:
```rust
    // SerialStream from tokio-serial crate is not Sync on Windows, so we need to spawn a separate thread for mmw_lib
    #[cfg(windows)]
    {
        // Create a new single-threaded runtime for Windows
        let runtime = tokio::runtime::Builder::new_current_thread()
            .enable_all()
            .build()
            .unwrap();

        // Spawn the radar connection handler on the new runtime
        std::thread::spawn(move || {
            runtime.block_on(async {
                radar_connection_handler(rx_radar_control).await;
            });
        });
    }

    #[cfg(not(windows))]
    {
        // Spawn a new tokio task to manage radar connection (original behavior for non-Windows platforms)
        tokio::spawn(async move {
            radar_connection_handler(rx_radar_control).await;
        });
    }
```

okhsunrog:
@sirhcel so the answer is: it can be only used in a single-threaded tokio runtime on Windows, sadly

unlimitedsola:
> I already added semver checks to CI with #240 but I'm apparently not entitled to rebase this PR on the current upstream master. @unlimitedsola, could you please rebase this PR on upstream/master or allow edits by maintainers?

Rebased, please take a look.

sirhcel:
Thank you for rebasing @unlimitedsola! The changes proposed here are semver-breaking - at least in theory.

Therefor I would prefer checking the our visible dependents from crates.io and GitHub whether they would be affected by this change first. I did this already in the past and have to dust off the makeshift setup used there. But this will take some time.

okhsunrog:
@sirhcel is there anything stopping it from being merged right now?

sirhcel:
> @sirhcel is there anything stopping it from being merged right now?

Yes, implementing `Sync` is actually a semver-breaking change. Cutting a release 5.0 just for this change seems a bit to much of semver compliance to me because I bet that this change does not affect depentents in practice. However, I want to give it a spin at least with our most-prominent dependents but I did not find the time for it the last weeks. ☹️

I don't know of tooling for this purpose (like a crater for libs) and used a [modified get-all-crates](https://github.com/sirhcel/get-all-crates/tree/dependents-filter) the last time. But I killed all the scripting I created back then when deleting the server. I need to find some time to set it up again. Any hints to prior art are welcome @okhsunrog.

sirhcel:
I picked the changes into [playground-5.0](https://github.com/serialport/serialport-rs/tree/playground-5.0).

unlimitedsola:
I think we can close this now as we have decided that we will adopt the approach in #327. Thanks for following up on this!

PR #236, opened 2024-12-20
Check for new issues from updated dependencies once a week

sirhcel:
We've experienced this more than once. Let's get aware of such issues early.


eldruin (approved):
Good idea, thanks!

PR #240, opened 2025-01-13
Check semver compatibility in CI

sirhcel:
This aims at automatically raising issues with semver compatibility. This is a quick and somewhat dirty shot to get the check in place at all. There is surely room for improvement.


eldruin (approved):
Looks good to me, thank you!

sirhcel:
This was quick :) thank you very much!

PR #242, opened 2025-01-14
Add note about DTR behavior on linux

jannic:
TL;DR: Linux always sets DTR on open, and you can't prevent that without kernel modifications.

See https://github.com/serialport/serialport-rs/issues/37 for more information, most importantly the linked stack overflow question. When working on #40 about 3 years ago, I verified that this was actually true. A quick search through the kernel git history didn't show any recent change.

The only thing I'm not sure about is if this is true for all serial ports on linux, or if it is driver dependent. So maybe the comment should be less strong? (Eg. "On Linux, DTR may automatically be set on open...")?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -396,6 +396,10 @@ impl SerialPortBuilder {
     }
 
     /// Set data terminal ready (DTR) to the given state when opening the device
+    ///
+    /// Note: On Linux, DTR is automatically set on open. Even if you set `dtr_on_open` to false,
+    /// DTR will be asserted for a short moment when opening the port. This can't be prevented
+    /// without kernel modifications.
     #[must_use]
     pub fn dtr_on_open(mut self, state: bool) -> Self {
         self.dtr_on_open = Some(state);


sirhcel:
Thank you for extending the documentiation @jannic! 

> The only thing I'm not sure about is if this is true for all serial ports on linux, or if it is driver dependent. So maybe the comment should be less strong? (Eg. "On Linux, DTR may automatically be set on open...")?

After a quick reading it looks to me like `serial_core.c` is initiating this and all driver supporting DTR will set it. Do you have a good reference from your earlier research on this topic? If not, the current note from this PR looks like a pretty good approximation on who sets DTR. I'm inclined to keep a low profile, let someone who really knows better prove us wrong and land this PR as-is.

jannic:
> Do you have a good reference from your earlier research on this topic?

Unfortunately no. Basically the only thing I have is the linked stack overflow question. This comment implies that it is possible to disable this automatic setting of DTR with a tiny kernel patch, on a driver by driver basis: https://unix.stackexchange.com/questions/446088/how-to-prevent-dtr-on-open-for-cdc-acm/543527#543527

So it is not impossible that some serial drivers do not implement the `dtr_rts` operation on `tty_port_operations` and therefore won't set DTR on open. Unfortunately, that's not easy to check: There definitely are `tty_port_operations` instances that don't implement this op, but not every ` tty_port_operations` belongs to a driver to a real serial port with a DTR line.

As far as I can tell, the comment in its current form is good for all practical purposes, even if not strictly true for some unconventional drivers.

Issue #248, opened 2025-02-25
"cargo test" fails and TESTING.md provides no guidance


JonathonReinhart:
[TESTING.md](https://github.com/serialport/serialport-rs/blob/main/TESTING.md) says:

> How to test `serialport-rs` for development.
>
> Without hardware:
>
>  1. Compilation
>  2. `cargo test`

---

When I try that it fails:

```
$ cargo test --no-default-features                                                                                                                                                           
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (target/debug/deps/serialport-083848d9b1c403f1)
                                                                                                                       
running 6 tests                                                                         
test posix::poll::tests::clamped_millis_c_int_is_monotonic ... ok                       
test posix::poll::tests::clamped_millis_c_int_zero_is_zero ... ok                       
test posix::poll::tests::clamped_time_spec_is_monotonic ... ok                        
test posix::poll::tests::clamped_time_spec_zero_is_zero ... ok                        
test posix::tty::test_ttyport_into_raw_fd ... ok                                      
test tests::timeout::tests::basic_durations_properties ... ok
                                                                                                                       
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/config.rs (target/debug/deps/config-ae02c05788e27e89)

running 0 tests                                                                       
                                                                                                                       
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
                                                                                                                       
     Running tests/test_baudrate.rs (target/debug/deps/test_baudrate-b7c732862a686999)
                                                                                                                       
running 27 tests                                                                                                 
test builder::test_non_standard_baud_rate_fails_where_not_supported::case_1 ... ignored      
test builder::test_non_standard_baud_rate_fails_where_not_supported::case_2 ... ignored
test builder::test_non_standard_baud_rate_fails_where_not_supported::case_3 ... ignored
test builder::test_non_standard_baud_rate_succeeds_where_supported::case_1 ... FAILED                            
test builder::test_non_standard_baud_rate_succeeds_where_supported::case_2 ... FAILED        
test builder::test_non_standard_baud_rate_succeeds_where_supported::case_3 ... FAILED
test builder::test_standard_baud_rate::case_1 ... FAILED 
test builder::test_standard_baud_rate::case_2 ... FAILED                            
[...]

PR #249, opened 2025-02-25
Fix testing features and docs

sirhcel:
Enabling hardware tests should be an additive feature and it should be documented in `TESTING.md`. This PR addresses #248.


sirhcel:
Yes, semver checks fail. But this is due to making the feature `ignore-hardware-tests` additive and renaming it to `hardware-tests`. I would opt to ignore this semver breaking change as this is a feature only affecting test execution.

eldruin (approved):
Sounds good to me. Thank you!

PR #255, opened 2025-03-25
Enumerate Bluetooth serial (RFCOMM) devices on Linux too

### LINKED ISSUE #246: On linux with udev, port whithout a parent should not be filtered out of `available_ports`
As discussed in #245, port whithout a parent value are filtered out durin enumeration with udev 

That should likely not be the case.

sirhcel:
* RFCOMM devices will be enumerated now (see #246)
* There are still issues with timeouts, ...
    * But they are independent from the enumeration
    * And some of them like baud rate setting are likely related to the nature of this connection type
* This PR also drive-by adds a test case for baud rates
* Fixes #246


sirhcel:
Semver checks are failing due to an unreleased semver breaking change from #249 considered not critical.

PR #254, opened 2025-03-24
Fix parsing serial numbers with underscores from Windows HWID

sirhcel:
* Fixes #253
* We've forgotten to at least include the underscore when replacing the former regex

PR #256, opened 2025-03-25
Release 4.7.1

sirhcel:
* The two low hanging fixes #254 and #255 have been integrated since then
* It has been some time since the last release
* Let's cut a patch release then
* A new release will get semver checks back on track as this release contains a semver-breaking change wit respect the feature for enabling/disabling hardware tests (and not relevant for normal use of this crate)


eldruin (approved):
Thank you!

PR #260, opened 2025-04-18
Allow compiling on tvOS, watchOS and visionOS too

madsmtm:
Assuming that the `mach2` crate is updated to support these platforms, or we get rid of that dependency.

I haven't tested this, only did the (very few) necessary changes to make it compile.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ nix = { version = "0.26", default-features = false, features = ["fs", "ioctl", "
 libudev = { version = "0.3.0", optional = true }
 unescaper = "0.1.3"
 
-[target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
+[target.'cfg(target_vendor = "apple")'.dependencies]
 core-foundation = "0.10.0"
 core-foundation-sys = "0.8.4"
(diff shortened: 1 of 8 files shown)


sirhcel:
Thank for for creating this PR @madsmtm!

> Assuming that the `mach2` crate is updated to support these platforms, or we get rid of that dependency.

Am reading it right, that you are aiming at a future point in time when support for tvOS, watchOS, and visionOS might get into the `mach2` crate? Do you know if there is already some work in this direction going on?

What do you mean with 'getting rid of that dependency'?

madsmtm:
> Do you know if there is already some work in this direction going on?

Unfortunately no, `mach2` seems unmaintained, and I don't know of a maintained alternative.

> What do you mean with 'getting rid of that dependency'?

The use of `mach2` in this crate is very minimal, basically two consts `KERN_SUCCESS` and `MACH_PORT_NULL` and a typedef `mach_port_t = c_uint`. So it'd be quite easy to get rid of entirely.

FYI, I'm somewhat planning to migrate this crate to `objc2`, see my [very in draft branch](https://github.com/madsmtm/serialport-rs/tree/objc2), I might do it as part of that.

PR #264, opened 2025-05-09
feat: Add integration with schemars

caspermeijn:
Integrate with `schemars`, which allows to generate a JSON schema for a `serde` generated JSON file.

A v1 of `schemars` is in the works, but that will take some months. It seems the derive API will not change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ features = [
 cfg-if = "1.0.0"
 scopeguard = "1.1"
 serde = { version = "1.0", features = ["derive"], optional = true }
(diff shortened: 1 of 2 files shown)


sirhcel:
Thank you for the PR @caspermeijn! What about adding support for schemars behind a preview feature for integrating it soon while waiting for serialport-rs to get to a suitable MSRV and maybe a breaking changes from schemars when transitioning to 1.0?

sirhcel:
I picked the changes into [playground-5.0](https://github.com/serialport/serialport-rs/tree/playground-5.0)

PR #266, opened 2025-05-14
unix: Use flock to ensure exclusive access

### LINKED ISSUE #265: Use flock for exclusive serial port access on Linux
This is a follow-up to #26, it seems that only `tiocexcl` is currently used, but not `flock`.

It would be nice to get support for `flock` as well, to catch more cases in which another application is already using the serial port.

I can create a PR adding support for this, if it something that would be accepted.

Tiwalun:
Some applications, e.g. [picocom](https://github.com/npat-efault/picocom), use `flock` to ensure exclusive access to the serial port, but don't seem to use the `tiocexcl` ioctl. This means that, without using `flock`, it is currently possible to connect to a port using picocom first, and then connect using this crate. This leads to problem where not all data is received by the Rust application.

To avoid this, `flock` can be used to acquire an advisory lock on the port. This will prevent this crate from opening the port when it has already been opened by picocom. Note that the lock is only advisory, meaning that this requires collaboration between applications.

Fixes #265.

### DIFF
--- a/src/posix/flock.rs
+++ b/src/posix/flock.rs
@@ -31,7 +31,7 @@ pub(crate) fn lock_shared(fd: RawFd) -> Result<()> {
         if e == nix::errno::Errno::EWOULDBLOCK {
             Error::new(
                 ErrorKind::NoDevice,
-                "Unable to acquire exclusive lock on serial port",
+                "Unable to acquire shared lock on serial port",
             )
         } else {
             e.into()


src/posix/tty.rs
```diff
@@ -130,6 +167,9 @@ impl TTYPort {
         // other applications that may have an exclusive port lock.
         ioctl::tiocexcl(fd.0)?;
 
+        // Also use flock to lock the port
+        flock_exclusive(fd.0)?;
```
sirhcel, line 171: After sleeping over it: Shouldn't we make this a "all or nothing" operation and attempt to gracefully retreat from this situation an not leave a port locked in the one way or the other?

sirhcel: Right. I should add a note on that for future me.

Tiwalun (commented):
Small comment in regards to the error message, but other than that it looks good.

eldruin (approved):
Great work, thank you!
Sounds like a good best-effort to me given the circumstances.
Regarding the `impl FromRawFd for TTYPort`, I think your solution is better that the status quote and strictly not worse, so it sounds good to me. Maybe we could have a fallible `TryFromRawFd` impl for this.

PR #271, opened 2025-05-20
windows overlapped io

erkki-silvola:


RossSmyth:
This PR has not comments anywhere, no explanation, and doesn't seem very well planned. I would prefer the other PR over this one. Thanks though.

sirhcel:
There is also #55 for the same topic.

erkki-silvola:
> There is also #55 for the same topic.

Yes, did look into to that originally. Compared to the other pr this solves the 0 timeout case, which is actually 0 compared to actually waiting for the event object where typical deviation is 4-20ms (when timeout > 0 there is is race between event object and timeout signal which is fine, though I guess if we want to rely on event object mostly then it could be so that if timeout > 0 then INFINITE is passed etc.), also somewhat cosmetic things related to handling event object(s) how it it normally done. My personal motivation would be to add later on cancellable rx event or eq. for streaming use case.

PR #278, opened 2025-07-29
Add event listener

TomzBench:
It looks like the only way to receive notifications of Usb plug and unplug events is by periodically calling "available_ports()" method.  We can avoid this polling by listening for Usb events using windows messaging on windows, and udev on unix.

This PR currently adds this support for windows. But if there is interest in this PR and the API is satisfactory then I can also add for linux.

The examples/listener output pasted below

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,10 +35,26 @@ features = [
     "minwindef", "ntdef", "setupapi", "winbase", "winerror", "winnt",
 ]
 
+[target."cfg(windows)".dependencies.windows-sys]
+version = "0.60"
+optional = true
+features = [
+	"Win32_Foundation",
+	"Win32_Graphics_Gdi",
+	"Win32_Security",
+	"Win32_System_Threading",
+	"Win32_System_LibraryLoader",
+	"Win32_System_Rpc",
+	"Win32_UI_WindowsAndMessaging",
+]
+
 [dependencies]
 cfg-if = "1.0.0"
 scopeguard = "1.1"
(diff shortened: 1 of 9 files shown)


TomzBench:
It can certainly be a separate crate. The additions basically do not touch any other files except to export the new methods. Though by introducing some coupling, we can potentially provide nicer API's down the road. And also I like the extra review due to the larger exposure this crate has.

If Maintenance is an issue, I use this crate and this listen feature professionally so have incentive to maintain it.

RossSmyth:
Yeah I use this crate professionally too and have become the unofficial windows person.

TomzBench:
Thanks for the comments. They are not off base at alll.

RE: 1-2 (no async) - I did not like adding the dependencies either... so I thought best to feature gated them. I guess the question is, does this repo intend to eventually support an async API?  I noted some PR's trying to add Overlapped IO, which made me think that somewhere down the line, async might be supported. 

If there is no ambitions to provide async api's down the road here, then it would further make sense to have this as a separate crate. Perhaps reference it in the readme here so that it might get some review benefits.

RE: 3 (message only window) I will do some tests with the message only windows. A brief view of the docs state that message only windows do not receive broadcasts messages. And device events are broadcast messages that must be explicitly enabled. So I'm not sure if it will work or not but will give it a try.

Speaking of windows messaging, the API currently is a little unfortunate that the user is required to pass the window name for their event listener. Maybe a randomly generated window name might be best to hide this detail, but also still support multiple listeners on the system.

Issue #281, opened 2025-08-13
Baudrate assertion fails in Linux


SergioGasquez:
In espflash we started encountering some issues when we change the baudrate: 
```
Error:   × Main thread panicked.
  ├─▶ at /home/sergio/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/serialport-4.7.2/src/posix/tty.rs:502:9
  ╰─▶ assertion failed: termios2.c_ospeed == termios2.c_ispeed
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

This seems to only happen in Linux (Im using ArchLinux) and happens with different espflash versions (I tested 3.3.0 and 4.0.1). See https://github.com/esp-rs/espflash/issues/938 for more details.

sirhcel:
> This seems to only happen in Linux (Im using ArchLinux) and happens with different espflash versions (I tested 3.3.0 and 4.0.1). See [esp-rs/espflash#938](https://github.com/esp-rs/espflash/issues/938) for more details.

Are you able to reproduce the issue with `TTYPort::baud_rate` with seralport's `hardware_check` example like

```
serialport-rs$ cargo run --example hardware_check -- /dev/ttyUSBn
```

where `ttyUSBn` is a ESP32 dev boart with a SiLabs CP2102N USB/UART bridge?

nickfarrow:
> Just in case the changed behavior from `ioctl::tcgets2` results from a recent change in the `cp210x` UART driver: Which version of the Linux kernel are you running?

@Lleny and I are both on `6.15.9-arch1-1` and I do suspect it is related to some recent kernel changes.

I was able to reproduce the issue via `hardware_check` using an ESP32 dev board with Espressif's built-in USB JTAG/serial debug unit (`USB ID 303a:1001`).

sirhcel:
> Explicitly passing a baud rate with `--baud 57600` does not change this. Do you have an idea how I could get espflash to actually call `TTYPort::set_baud`? I'm pretty sure that my instrumentation is actually active as putting a `panic!` next to it, the compiler warns me about unreachable statements below the panic.

I finally managed to get `TTYPort::baud_rate` called from espflash with `--baud 921600` @SergioGasquez. My instrumentation shows the baud rates to be indentical in my case:

```
espflash/espflash$ cargo run --release -- flash --baud 921600 --monitor /tmp/main
[...]
[...]

PR #283, opened 2025-08-17
fix input baudrate not cleared

33671:
```rust
pub(crate) fn set_baud_rate(termios: &mut Termios, baud_rate: u32) -> Result<()> {
    termios.c_cflag &= !nix::libc::CBAUD;
    termios.c_cflag |= nix::libc::BOTHER;
    termios.c_ispeed = baud_rate;
    termios.c_ospeed = baud_rate;
}
```
`termios.c_cflag &= !nix::libc::CBAUD;` only cleared CBAUD but left CIBAUD untouched. since the termios struct passed in is the one previously read from the kernel, it already contains whatever input baud rate flags were set earlier.

If we don’t clear them, the old CIBAUD sticks around while ospeed gets updated, causing a mismatch: 
```bash
assertion failed: termios2.c_ospeed == termios2.c_ispeed
```
see issue: https://github.com/serialport/serialport-rs/issues/281

This is consistent with glibc implementation (https://github.com/bminor/glibc/blob/35296a6e73d3244b415823e2811fa5f930457944/sysdeps/unix/sysv/linux/cfsetspeed.c#L29)

I verified this change on my own virtual serialport module: https://github.com/33671/vserial/, and it fixed the baudrate mismatch there.
espflash is not tested, but given the similarity to glibc’s approach, this should resolve the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ project adheres to [Semantic Versioning](https://semver.org/).
   [#29](https://github.com/serialport/serialport-rs/pull/29)
 
 ### Fixed
+
+* Setting arbitrary baud rates on Linux which resulted in issues when read back
+  on Arch recently.
+  [#281](https://github.com/serialport/serialport-rs/issues/281)
+  [#283](https://github.com/serialport/serialport-rs/pull/283)
+
 ### Removed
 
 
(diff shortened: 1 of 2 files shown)


33671:
Yes, your are right.the current code basically assumes the kernel will handle termios setting correctly, there may be other potential problems for example, it always sets BOTHER in c_cflags instead of using the standard baud rate flags.

I kept this PR minimal，whether to move to nix’s safe functions (which may require refactoring most of termios.rs) or to align with glibc’s approach is up to you maintainers.🙂

Issue #277, opened 2025-07-21
Allow non-blocking behavior


Gip-Gip:
I have a library which is dependent on using timing to determine the end of a packet. Unfortunately, when it times out it A: throws an error which is opaque to my library, since it uses `embedded_io` traits which are assumed to be non-blocking(and doesn't convert error types), and B: times out sooner than I would expect. I have the timeout duration set to 1 ms and it receives all of the data under 500us but still the Linux driver or something just times out and I get no data.

I believe it would be beneficial to have an option for the SerialPortBuilder to allow for non-blocking reading, as it would allow for my code to have lower latency. That and when interfacing with other embedded libraries, like Modbus for example, they expect non-blocking I/O.

robamu:
I think this would be useful as well.
Maybe there is a smarter way, but that would avoid a wrapper like this:

```rust
    /// Thin wrapper around [std::io::Read] which converts calls which do not return anything
    /// to Ok(0) to allow polled reads.
    pub fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        match self.port.read(buf) {
            Ok(read_bytes) => Ok(read_bytes),
            Err(e) => {
                if e.kind() != std::io::ErrorKind::TimedOut
                    && e.kind() != std::io::ErrorKind::WouldBlock
                {
                    return Err(e);
                }
                Ok(0)
            }
        }
    }

```

PR #289, opened 2025-08-29
Added new VMIN/VTIME API for TTYPort

### LINKED ISSUE #272: Support configuration of VMIN and VTIME
Configuration of how long to block for (in terms of number of characters) doesn't appear to be supported  currently. Consider adding support in the POSIX apis for VMIN and VTIME, or some other abstraction that essentially does that.
http://www.unixwiz.net/techtips/termios-vmin-vtime.html

robamu:
I still need to test this with some actual hardware (it might not work at all, I am still new to nix/termios etc. with Rust :))  but this possibly fixes #272 and #277 .

- Add `set_vmin_time` method to configure VMIN and VTIME termios parameters
- Add `set_non_blocking` method to set VMIN and VTIME to 0


robamu:
I implemented something more generic, but I am not fully sure how that interacts with the `poll`  API used. I guess I'll find out during testing...

Issue #231, opened 2024-12-02
Will rust 1.59.0 be a barrier to using the newer create?


sweet-de-way:
Over time, many third-party libraries are updated to take advantage of new features and improvements in Rust. If your project depends on these libraries and they no longer support Rust 1.59.0, using an older version of the compiler may become a problem.

For example, windows-sys 0.59.0 requires libc 0.2.164 to build on some platforms.

#228

sirhcel:
Yes, using MSRV 1.59.0 is already a barrier and this is getting more of a challenge as rust releases progress. The reason we are using this "prehistoric" release as our MSRV is the latest long-term support release of Yocto (for building embedded Linux distributions) provided this Rust release.

A new long-term support release of Yocto is out and I wanted to tackle this long overdue bump of the MSRV together with a bigger refactoring of the interface (like #34, #94 ...). But it looks like this is too much asked and not gonna happen soon.

I consider bumping our MSRV as a semver breaking change. The seralport crate stumbled more than one time over dependencies bumping their MSRV with a minor release and I'd like to spare users this not so pleasant experience. That's why I'm starting to get inclined to do split this up in two major releases:

* Release 5.0 for bumping the MSRV to 1.75.0 (which is the one supported by the current long-term support release of Yocto)
* Releaso 6.0 for transitioning to a platform-specific representation of `SerialPort`

So a switch to windows-sys could land with each of them.

unlimitedsola:
> * Release 5.0 for bumping the MSRV to 1.75.0 (which is the one supported by the current long-term support release of Yocto)

Hi, I'm also seeing a lot of PRs are currently blocked by this MSRV requirement, so I wonder when could this happen? Should we start branching out with a newer MSRV so that we can start getting these PRs merged?

sirhcel:
> Hi, I'm also seeing a lot of PRs are currently blocked by this MSRV requirement, so I wonder when could this happen? Should we start branching out with a newer MSRV so that we can start getting PRs merged and have a foundation to work on?

Which PRs are you referring to here? For sure, the list of PRs and also tagging them. As of now both, semver breaking and MSRV breaking changes are tagged as _5.0.0_. We could and likely should split this up between semver and MSRV breaking changes.

unlimitedsola:
> Which PRs are you referring to here?

[...]

PR #296, opened 2025-10-05
Use `objc2-io-kit`

madsmtm:
Use [`objc2-io-kit`](https://docs.rs/objc2-io-kit/) and [`objc2-core-foundation`](https://docs.rs/objc2-core-foundation/) instead of [`io-kit-sys`](https://github.com/jtakakura/io-kit-rs) and [`core-foundation`](https://docs.rs/core-foundation/).

These are part of [the `objc2` project](https://github.com/madsmtm/objc2) which I maintain, which provides auto-generated bindings for Apple's APIs. This allows us to be much more confident that the bindings are correct, handles memory management like `CFRetain`/`CFRelease` automatically, allows us to do cool things like embedding documentation from the headers.

Note also that `core-foundation` has been soft-deprecated in favour of `objc2-core-foundation`, see https://github.com/servo/core-foundation-rs/issues/729, so it makes sense to start this migration.

This builds upon https://github.com/serialport/serialport-rs/pull/260, so that should probably be merged first. This change also requires an MSRV bump to 1.71 because these crates use `extern "C-unwind"`, so this should probably be put in the `v5.0.0` milestone IIUC, see also https://github.com/serialport/serialport-rs/issues/231.

### DIFF
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -8,7 +8,7 @@ cfg_if! {
 
 cfg_if! {
     if #[cfg(target_vendor = "apple")] {
-        use core::ffi::{c_char, c_int, c_uint, CStr};
+        use core::ffi::{c_int, c_uint, CStr};
         use core::mem::MaybeUninit;
 
         use objc2_core_foundation::{
@@ -325,7 +325,7 @@ fn get_parent_device_by_type(
 ) -> Option<io_registry_entry_t> {
     let mut device = device;
     loop {
-        let mut class_name = MaybeUninit::<[c_char; 128]>::uninit();
+        let mut class_name = MaybeUninit::uninit();
         unsafe { IOObjectGetClass(device, class_name.as_mut_ptr()) };
         let class_name = unsafe { class_name.assume_init() };
         let name = unsafe { CStr::from_ptr(&class_name[0]) };
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -392,6 +392,8 @@ fn get_int_property(device_type: io_registry_entry_t, property: &str) -> Result<
 fn get_string_property(device_type: io_registry_entry_t, property: &str) -> Result<String> {
     let cf_property = CFString::from_str(property);
 
+    // SAFETY: We are calling `IORegistryEntryCreateCFProperty` with valid arguments and have
+    // allocated `cf_property` right above.
     let cf_type = unsafe {
         IORegistryEntryCreateCFProperty(device_type, Some(&cf_property), kCFAllocatorDefault, 0)
     }
--- a/src/posix/enumerate.rs
+++ b/src/posix/enumerate.rs
@@ -8,20 +8,21 @@ cfg_if! {
 
 cfg_if! {
     if #[cfg(target_vendor = "apple")] {
-        use core_foundation::base::CFType;
-        use core_foundation::base::TCFType;
-        use core_foundation::dictionary::CFDictionary;
-        use core_foundation::dictionary::CFMutableDictionary;
-        use core_foundation::number::CFNumber;
-        use core_foundation::string::CFString;
-        use core_foundation_sys::base::{kCFAllocatorDefault, CFRetain};
-        use io_kit_sys::*;
-        use io_kit_sys::keys::*;
-        use io_kit_sys::serial::keys::*;
-        use io_kit_sys::types::*;
-        use io_kit_sys::usb::lib::*;
-        use core::ffi::{c_char, c_int, c_uint, c_void, CStr};
+        use core::ffi::{c_char, c_int, c_uint, CStr};
         use core::mem::MaybeUninit;
+
+        use objc2_core_foundation::{
+            kCFAllocatorDefault, CFMutableDictionary, CFNumber, CFRetained, CFString, CFType,
+        };
+        #[allow(deprecated)]
+        use objc2_io_kit::{kIOMasterPortDefault, IOMasterPort};
+        use objc2_io_kit::{
+            io_object_t, io_registry_entry_t, kIOServiceClass, kIOUSBDeviceClassName,
+            kIOUSBHostInterfaceClassName, IOIteratorNext, IOObjectGetClass, IOObjectRelease,
+            IORegistryEntryCreateCFProperties, IORegistryEntryCreateCFProperty,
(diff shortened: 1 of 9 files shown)


eldruin (commented):
Thank you for this work and your work on the `objc2` crates!
This sounds good to me.

FYI:
Yocto kirkstone (Rust 1.59.0) goes into EOL on April 2026.
The next supported release is Scathgap, which will have Rust 1.75.0.

sirhcel:
Hello @madsmtm, thank you very much for preparing this PR! I would like to integrate it into [playground-5.0](https://github.com/serialport/serialport-rs/tree/playground-5.0) for working towards our next major release. I did not spot an option for changing the target branch of this PR. But I bet you can. Could you please change it?

I would also rebase the changes onto the current playground-5.0 branch and as I have done or followed the edits there, you can find my conflict resolution in [sirhcel@test/objc2](https://github.com/sirhcel/serialport-rs/tree/test/objc2).

madsmtm:
I think I've done so now

sirhcel:
Thank you very much @madsmtm! I rebased the changes on the current playground and started with some nitpicking. I found out about the Matrix space from https://github.com/madsmtm/objc2#contact and will ask some more general questions over there.

sirhcel:
Thank you very much for answering all my loosely related questions in the chat! I feel i'm done with my nitpicking. Also addressing `available_ports` with respect to the existing use of unsafe in this pr seems to much for the moment to me and I leave this to #356.

Could you please give my nitpicking and cleanup a review @madsmtm? I feel ready to integrate this PR and we should leave the points discussed in the chat for another round.

src/posix/enumerate.rs
```diff
@@ -274,23 +272,36 @@ fn get_parent_device_by_type(
 ) -> Option<io_registry_entry_t> {
     let mut device = device;
     loop {
-        let mut class_name = MaybeUninit::<[c_char; 128]>::uninit();
-        unsafe { IOObjectGetClass(device, class_name.as_mut_ptr()) };
+        let mut class_name = MaybeUninit::zeroed();
```
madsmtm, line 328: I approve of using `zeroed` over `uninit`, there's very little documentation in Apple's APIs that they don't read stuff.

src/posix/enumerate.rs
```diff
@@ -274,23 +272,36 @@ fn get_parent_device_by_type(
...
+        // SAFETY: `device` is supposed to be valid `io_object_t` and `class_name` points to a
+        // zero-initialzed buffer of the required size.
```
madsmtm, line 330: From my understanding, it's not a safety requirement that the `device` (and `io_object_t`s in general) is valid? It's an ID, not a pointer, and if I call `IOObjectGetClass(42, &mut class_name)`, the validity of the device will be checked by the kernel and I'll just get an error, not UB.

src/posix/enumerate.rs
madsmtm: Though if you're zero-initializing, you could make it even safer by just using an array:
```rust
let mut class_name = [0; 128];
if KERN_SUCCESS != unsafe { IOObjectGetClass(device, &mut class_name) } {
    return None;
}
let name = CStr::from_bytes_until_nul(&class_name).unwrap();
```

src/posix/enumerate.rs
madsmtm: (If you agree, then I'll probably be able to mark a lot of IOKit APIs as safe).

src/posix/enumerate.rs
```diff
@@ -402,7 +419,7 @@ cfg_if! {
                 // Run the search.
                 let mut matching_services = MaybeUninit::uninit();
                 kern_result = IOServiceGetMatchingServices(
-                    kIOMasterPortDefault,
+                    kIOMainPortDefault,
```
madsmtm: This will raise your deployment target. I'd recommend that you keep using `kIOMasterPortDefault` and `IOMasterPort`, and just keep using `#[allow(deprecated)]`.

See also https://github.com/madsmtm/objc2/issues/824 and https://github.com/GuillaumeGomez/sysinfo/pull/1641.

sirhcel: Thank you! I was not aware of this. I will leave it as-is then.

src/posix/enumerate.rs
sirhcel: I looked into using an array but the types do not play out for using [`CStr::from_bytes_until_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul): This method expects `&[u8]` where dereferencing the array `&[i8]` aka `&[c_char]`. I find this interesting as [`CStr::from_ptr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr) takes `*const i8`. This looks like we're facing some "impedance mismatch" between IO-Kit and `CStr`.

At a first glance I would prefer sticking to the current implementation in favor of getting around casting pointers with `as`. Do you see a better route from `[c_char; 128]` to `CStr`?

src/posix/enumerate.rs
sirhcel: I see the point you are making with `io_object_t` being an ID/handle. I have to admit that I have not much practical experience with using Apples APIs beyond serialport-rs. And ever after researching documentation from Apple on that, I did not find any statement on that issue. So i don't feel confident to deem this use case safe. Do you have any references on that for me?

src/posix/enumerate.rs
madsmtm: Ah, you're right

> Do you see a better route from `[c_char; 128]` to `CStr`?

Not really, no.

src/posix/enumerate.rs
madsmtm: Not really any other reference than "`IOObjectRetain(42)` does not crash, while `CFRetain(42 as *const void)` definitely does" and "this is what I'd expect from kernel-backed objects (similar to file handles)".

(Also the only reason why those retain/release functions would return a status code).

(Actually, looking at it, it seems I've already marked these things safe in https://github.com/madsmtm/objc2/commit/0145877b8a5058a92b4e32fdbdeb68c43bf02362, and I think I stand by that reasoning).

src/posix/enumerate.rs
madsmtm: Wait, actually yes: You can do:
```rust
let name = class_name.map(|c| c as u8);
let name = CStr::from_bytes_until_nul(&class_name).unwrap();
```

PR #280, opened 2025-08-06
Replace `winapi` with `windows-sys` crate

unlimitedsola:
This is another take of https://github.com/serialport/serialport-rs/pull/228.

- Fixes <https://github.com/serialport/serialport-rs/issues/56>
- Closes <https://github.com/serialport/serialport-rs/pull/228>

This PR offers two options:

- [`windows-sys 0.62`](https://crates.io/crates/windows-sys/0.60.2) (requires MSRV bump to 1.60) if only take the first commit
- [`windows-sys 0.52`](https://crates.io/crates/windows-sys/0.52.0) (MSRV 1.52) with both commits

Should you choose not to squash the commits when merging, you can then revert the second commit later when you decide to upgrade the `windows-sys` crate (as upgrading introduces breaking changes) to make it compatible with newer version.

Test results:

- `windows-sys 0.62`: <https://github.com/sola-contrib/serialport-rs/actions/runs/16767971224>
- `windows-sys 0.52`: <https://github.com/sola-contrib/serialport-rs/actions/runs/16769159151>

PR #297, opened 2025-10-06
Release 4.8

sirhcel:
With three PRs landet since the last release

* [#266]
* [#280]
* [#290]

it's a good point in time to cut the next one.


sirhcel:
The failing builds for checking MSRV compatibility are caused by a recent update to our sub-dependency quote (through serde-derive). This dependency gets pinned with #295 which will resolve the issue.

eldruin (approved):
Sounds good to me, thank you!
I incorporated the fixes from #295 via rebase, so please update your local copy before publishing.

PR #299, opened 2025-10-07
Fix gerenating documentation at docs.rs and release 4.8.1

sirhcel:
Generating the latest documentation on docs.rs failed with https://docs.rs/crate/serialport/4.8.0/builds/2566488. It looks like https://github.com/rust-lang/rust/pull/138907 landed there and we have to update our configuration for it.


sirhcel:
Thank you for the quick review @eldruin!

Issue #302, opened 2025-10-16
Tracking issue for release 5


sirhcel:
Bumping our MSRV is overdue. A fair amount of issues is waiting for it to get integrated. Let's sort what brings semver breaking changes and what requires a higher MSRV just for getting integrated. The milestone https://github.com/serialport/serialport-rs/milestone/1 does not make this distinction yet.

So let's get this planned at all and have a place for the discussions. Let's get started at all and update this issue as we go. Your Feedback and contributions are welcome.

Ideas:

* [ ] Switch from different platform-specific implementations of the `TTYPort` trait to a single struct (with platform specific implementations)
* [ ] Provive platform-specific functionality via extension traits
    * Like controlling RS-485 port settings on Linux (#226)
    * #330
* [ ] Allow splitting a serialport instance into a `io::Read` and `io::Write` half to conveniently allow reading and writing from different threads without synchronization overhead

PRs to integrate bringing semver breaking changes:

* [ ] #195
* [ ] #189
* [ ] #233 

PRs requiring a higher MSRV for integration:

* [ ] #263 
* [ ] #264 

Issues and PRs requiring some discussion and weighing: 

* [ ] #317

ColinFinck:
@sirhcel: Thank you for your quick reply!

> Are you sure with that version number? Looking at https://crates.io/crates/windows-sys/versions shows me 0.61.2 at the latest released version. I bumped windows-sys in the plaground to this version and it builds afters some minor modifications.

I'm sorry, I was looking at the version tags in the https://github.com/microsoft/windows-rs repo and didn't realize that these version numbers weren't synchronized with each sub-crate... 🙈 
https://github.com/microsoft/windows-rs/pull/3741 states that "windows-rs Release 69" contains `windows-sys 0.61.0`, `windows-link 0.2.0`, etc.. 🙈

sirhcel:
> What's left to do for point 1? Would you mind opening a tracking issue that lists all the things that you would want to see in such an MSRV bump major release? I might find some time to support you with those, once split into discrete tasks.

Thank you very much for your offer @Finomnis! I rushed that through with #328 and cut release 4.9.0. I hope this resolves the MSRV issue for you as well.

sirhcel:
> I experience issues with using nusb and serialport-red at the same time.

Does release 4.9.0 resolve the issue for you @BjornTheProgrammer?

sirhcel:
[...]

PR #303, opened 2025-10-25
Added FromStr impls for enums that have Display trait

TheBeardedQuack:
Many enums in the crate provide the `Display` trait, while very few provide `FromStr`.
The `FlowControl` enum for example has been provided with both traits, while `DataBits`, `StopBits` and `Partity` enums feature only the `Display` trait.

Added `FromStr` imlementation to these three enums to improve completeness and assist in applications that require converting from a string in the UI into these enum values.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,7 +192,7 @@ impl From<DataBits> for u8 {
 impl TryFrom<u8> for DataBits {
     type Error = ();
 
-    fn try_from(value: u8) -> core::result::Result<Self, Self::Error> {
+    fn try_from(value: u8) -> std::result::Result<Self, Self::Error> {
         match value {
             5 => Ok(Self::Five),
             6 => Ok(Self::Six),
@@ -294,7 +294,7 @@ impl From<StopBits> for u8 {
 impl TryFrom<u8> for StopBits {
     type Error = ();
 
-    fn try_from(value: u8) -> core::result::Result<Self, Self::Error> {
+    fn try_from(value: u8) -> std::result::Result<Self, Self::Error> {
         match value {
             1 => Ok(Self::One),
             2 => Ok(Self::Two),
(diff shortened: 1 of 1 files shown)


sirhcel:
Thank you for spotting this discrepancy @TheBeardedQuack and preparing a PR! Are you actively using one of the new  `FromStr` implementations for the types above? I'm asking because this PR is one of the few occasions this topic was brought on the table and i am assuming that only few people are actively using it.

On the other hand there is also the argument of symmetry of `Display` and `FromStr` according to [Completeness and parseability](https://github.com/serialport/serialport-rs/pull/303#issuecomment-3448768706) from `Display` which - putting the localization issue aside - perfectly matches how we implement `Display` here.

I also see your point with `Display` @RossSmyth. Display is already on board here for quite some time. Removing will be a semver breaking change so this is also a topic for #302. May be we can deprecate the implementations of `Display` and provide them through `Debug`. From my point of view, something `FromStr`-ish is useful. I'm inclined to land #264 for 5.0 and with this, having a way to parse a textual representation without schema/JSON looks like a fair thing to offer at a first glance. I added #317 (and linked it to #302) to continue the discussion about implementing `Display` there.

sirhcel:
After reading through [`FromStr`s Input format and round-tripping](https://doc.rust-lang.org/std/str/trait.FromStr.html#input-format-and-round-tripping) for several times, it looks to me that there is no contract preventing us from having round-tripping for `Debug` and `FromStr` while not implementing `Display` at all.

From my point of view this allows us to add a `FromStr` allowing round-tripping with `Debug` and `Display` now and decide on the future of `Display` separately (#317, declaring it a "[developer log locale](https://doc.rust-lang.org/std/fmt/trait.Display.html#internationalization)", or dropping it). And I follow @TheBeardedQuack's argument that there is value in having a textual representation which can be generated and parsed without serde.

I still see two points open for this PR:

1. Adding a tests for round-tripping (with `Debug` and `Display` in the first place)
2. Checking if there are other meaningful cases we should cover with `FromStr` as well. I'm thinking about parsing `N`, `E`, and `O` for `Parity` (in upper and lower case) as this is something someone could easily want to process as an input for `Parity::from_str` when dealing with a textual representation like `115200 8N1` which seems common to me.

I will give the former a spin in the next days.

sirhcel:
Rebased onto current main to get rid of unnecessary noise from CI.

PR #316, opened 2026-02-21
Enable arbitrary baud rates on Linux musl

jameshilliard:
libc now exposes termios2, TCGETS2, TCSETS2, and BOTHER on musl. Use the same termios2-based arbitrary baud-rate handling there as on glibc targets.

Remove target_env = "musl" from cfg exclusions and keep the legacy termios fallback only for powerpc and powerpc64.

Update tests that treated linux+musl as the fallback path so they key off powerpc and powerpc64 instead, and document the fix in the changelog.

### DIFF
--- a/tests/test_baudrate.rs
+++ b/tests/test_baudrate.rs
@@ -138,7 +138,10 @@ mod builder {
     #[cfg_attr(
         any(
             not(feature = "hardware-tests"),
-            all(target_os = "linux", target_env = "musl"),
+            all(
+                target_os = "linux",
+                any(target_arch = "powerpc", target_arch = "powerpc64")
+            ),
         ),
         ignore
     )]
@@ -230,7 +233,10 @@ mod new {
     #[cfg_attr(
         any(
             not(feature = "hardware-tests"),
-            all(target_os = "linux", target_env = "musl"),
+            all(
+                target_os = "linux",
+                any(target_arch = "powerpc", target_arch = "powerpc64")
+            ),
         ),
         ignore
     )]
@@ -339,7 +345,10 @@ mod set_baud {
     #[cfg_attr(
         any(
             not(feature = "hardware-tests"),
-            all(target_os = "linux", target_env = "musl"),
+            all(
+                target_os = "linux",
+                any(target_arch = "powerpc", target_arch = "powerpc64")
+            ),
         ),
         ignore
     )]


sirhcel:
Thank you for bringing this up and this PR! Good catch! Builds and the tests run locally just fine.

With _libc_, you are referring to the libc crate right? Do you know when the change with respect to musl happened there? Might this be already years back? I'm asking because at a first glance I did not spot a change related to the termios2 constants or their exposure with respect to musl.

jameshilliard:
> Thank you for this reference! It looks to me that the constants were made publicly avalilable with this commit but I don't see a distinction between musl and other libc implementations in this commit and at all. Do i miss something obvious here? Or was such a distinction just never made by the libc crate?

I'm guessing without that change the constants were getting picked up somehow from glibc but not musl due to not being present in normal glibc but not musl. I'm not entirely sure how that rust libc wrapper works entirely however.

> But without further comments on the cause and so the reason for the distinction on our side is still not obvious to me.

I'm assuming that would have been due to the missing constants or something along those lines.

sirhcel:
Thank you for opening up this rabbit hole @jameshilliard! And this is meant in a positive way. I rebased your PR to slip in a missing test for non-standard baud rates and cross-checked the actual baud rates on the wire - and they match what the test requests.

> I'm guessing without that change the constants were getting picked up somehow from glibc but not musl due to not being present in normal glibc but not musl. I'm not entirely sure how that rust libc wrapper works entirely however.

It's getting really interesting as things are not as I expected them to be in the first place. It looks like we are bypassing the system libc entirely:

* When switching to using the `termios2` interface with Linux/musl , the termios support from musl does not come into play and serialport [defines and uses](https://github.com/serialport/serialport-rs/blob/98e5eeb8698756e6af2055848b8d423205389338/src/posix/ioctl.rs#L66) its own ioctl wrappers
* The `termios2` struct is [defined by the libc crate](https://docs.rs/libc/0.2.180/src/libc/unix/linux_like/linux/arch/generic/mod.rs.html#4-15) on its own and not generated from the actual system libc in use

This explains to me how this works out phenomenologically without having any apparent support for `termios2` in musl: it is done by directly talking to the Linux kernel via the ioctl interface. Wild!

For me the question now is whether this is a sustainable way of enabling non-standard baud rates for musl. My first reaction is 'why not' because the ioctls are specified and just business between serialport and the Kernel itself then. Let me "sleep" over it.


> > But without further comments on the cause and so the reason for the distinction on our side is still not obvious to me.
> 
> I'm assuming that would have been due to the missing constants or something along those lines.

Maybe. This would require looking more into ancient releases of libc and nix to check how things were back then at the time of 869f095.

jameshilliard:
> For me the question now is whether this is a sustainable way of enabling non-standard baud rates for musl. My first reaction is 'why not' because the ioctls are specified and just business between serialport and the Kernel itself then.

Yeah, presumably this is one of the reasons rust's libc defines these sort of things.

jameshilliard:
> I gave it a scan, I don't see anything that would be internal bookkeeping except for setting errno.

Yeah, I think there isn't really all that much special going on with termios2, these operations are fairly straightforward syscalls AFAIU.

PR #318, opened 2026-02-22
Add simple data transmission test fon non-standard baud rates

sirhcel:
This test should have been added along with the data transmission at standard baud rates back then.


sirhcel:
Stumbled upon this while looking at #316

PR #320, opened 2026-02-27
Add configurable ReadIntervalTimeout for Windows COMPort #275

### LINKED ISSUE #275: Windows configurable ReadIntervalTimeout
Hi

I would much like some way of configuring win32 ReadIntervalTimeout.

My application uses a fork of `serialport-rs:4.2.1` with modified `ReadIntervalTimeout`  to `1` (instead of `0` at that time, which was extrely slow as I recall).

I noticed that you have since changed `ReadIntervalTimeout` to `MAXDWORD`, so I retested:

`serialport-rs:4.2.1` with modified `ReadIntervalTimeout`  to `1` :
transfer rate 31.5 kbytes/s

`serialport-rs:4.7.2`
transfer rate 20.5 kbytes/s

(FYI: Both tests run with registry setting `"SYSTEM\CurrentControlSet\Enum\FTDIBUS\VID_nnnn+PID_nnnn+nn\0000\Device Parameters"` = `1`)

VMshall:
This PR resolves #275 by exposing the `ReadIntervalTimeout` configuration natively for Windows `COMPort` structs. 

* Added `read_interval_timeout: u32` state tracking to `COMPort`.
* Added `pub fn set_read_interval_timeout(&mut self, timeout: u32) -> Result<()>` which updates the tracked state and calls `self.set_timeout` to push the new `COMMTIMEOUTS` payload down to the OS.
* Added a getter `pub fn read_interval_timeout(&self) -> u32`.
* Ensure `try_clone_native()` propagates the requested timeout value to cloned handles.
* Default constructor behavior (`u32::MAX`) remains completely untouched to avoid breaking downstream reliance on the default non-blocking behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,8 @@
 //! platform-specific port object which allows access to platform-specific functionality.
 
 #![allow(clippy::uninlined_format_args)]
+// Clippy's MSRV lint can disagree with macro-expanded code from dependencies (e.g. nix ioctl helpers) even when `cargo build` at the crate MSRV still succeeds. See CI `clippy` jobs.
+#![allow(clippy::incompatible_msrv)]
 #![deny(
     clippy::dbg_macro,
     missing_docs,
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -110,19 +110,19 @@ impl COMPort {
 
     /// Set the `ReadIntervalTimeout` in milliseconds
     ///
-    /// `ReadIntervalTimeout` is the maximum time allowed to elapse before the arrival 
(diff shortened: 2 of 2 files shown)


VMshall:
Hi @NONnonHere , thank you for catching this! Your interpretation of the Win32 `COMMTIMEOUTS `documentation is spot on.

You're absolutely right that if `ReadIntervalTimeout` is set to anything other than `MAXDWORD (u32::MAX)`, leaving `ReadTotalTimeoutMultiplier` at `MAXDWORD` will overflow the total timeout calculation and break the strict semantics established in #79.

The dynamic fix you suggested for `ReadTotalTimeoutMultiplier` makes perfect sense.

PR #327, opened 2026-03-12
Use OwnedHandle on Windows

unlimitedsola:
This is the alternative approach mentioned in https://github.com/serialport/serialport-rs/pull/233. The code is a bit more verbose due to `windows-sys` APIs require raw handles. This is provided as an option, feel free to close this if you feel the verbosity is not worth the removed `unsafe impl`.

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -131,9 +131,9 @@ impl COMPort {
                 DUPLICATE_SAME_ACCESS,
             )
         };
-        if duplicated && cloned_handle != INVALID_HANDLE_VALUE {
+        if duplicated == 0 && cloned_handle != INVALID_HANDLE_VALUE {
             Ok(COMPort {
-                handle: OwnedHandle::from_raw_handle(cloned_handle),
+                handle: unsafe { OwnedHandle::from_raw_handle(cloned_handle) },
                 port_name: self.port_name.clone(),
                 timeout: self.timeout,
             })
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -119,7 +119,8 @@ impl COMPort {
     pub fn try_clone_native(&self) -> Result<COMPort> {
         let process_handle: HANDLE = unsafe { GetCurrentProcess() };
         let mut cloned_handle: HANDLE = INVALID_HANDLE_VALUE;
-        unsafe {
+
+        let duplicated = unsafe {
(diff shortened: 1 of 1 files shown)


sirhcel:
Thank you for this follow-up! At a first glance I don't mind having the more verbose `.as_raw_handle()` all over the place in favor of not having to discuss why manually implementing `Send` and `Sync` is save for us to do.

src/windows/com.rs
```diff
@@ -135,7 +131,7 @@ impl COMPort {
             );
             if cloned_handle != INVALID_HANDLE_VALUE {
                 Ok(COMPort {
-                    handle: cloned_handle,
+                    handle: OwnedHandle::from_raw_handle(cloned_handle as RawHandle),
```
sirhcel: I don't see the need for casting here at a first glance. Isn't `cloned_handle` already a `RawHandle`?


```suggestion
                    handle: OwnedHandle::from_raw_handle(cloned_handle),
```

src/windows/com.rs
```diff
@@ -135,7 +131,7 @@ impl COMPort {
             );
             if cloned_handle != INVALID_HANDLE_VALUE {
```
sirhcel: Well, this is blast from the past. But as far as I'm overseeing things here we were ignoring the result from `DuplicateHandle`. The error handling below already fetches the cause behind the curtain by `last_os_error` with `GetLastError`. So I would add this check while at it. I can't update this PR or make a suggestion for this piece of code. Could you please cherry-pick [f45f839](https://github.com/sirhcel/serialport-rs/commit/f45f83984f5e3655d07865e6b87b2496a5a3ddf0)?

sirhcel (dismissed):

src/windows/com.rs
unlimitedsola: Done! Please take another look.

src/windows/com.rs
unlimitedsola: Good catch! I've removed both casts to `RawHandle`.

src/windows/com.rs
sirhcel: Thank you! I'm sorry for messing my commit up! I wanted it to be `duplicated == TRUE` and will fix this after integrating.

sirhcel:
Thank you very much for this contribution!

Issue #324, opened 2026-03-10
Please stop pinning dependency versions.


Dirbaio:
`serialport` pins maximum versions of dependencies in `Cargo.toml` to prevent Cargo from picking dependency versions that have a MSRV higher than `serialport`'s MSRV.

This is a problem because it prevents *everyone* from using `serialport` together with the newer dependency versions, even if they're using a newer Rust version.

This is a problem because:
1. It prevents upgrading dependencies, locking users to old versions that have known bugs and potential security issues.
2. It can completely break the build if using another library that requires a newer dependency version. Cargo will refuse to compile since it won't allow compiling two different minor versions of the shared deps and the acceptable version ranges of `serialport` and the other lib don't overlap.

This is a recurring pain I run into basically every single project using `serialport-rs`, even if indirectly e.g. through `probe-rs`. It's so recurring that at work we have a fork of this repo just to remove the version pins. Examples of other users running into this issue: #307 #304 #300 #231

The solution is to ensure `serialport` *can* build with old dependency versions that match the MSRV, but not *prevent* building with newer dependency versions if the user wishes to do that. To do this:

- Don't pin dependency versions with `=` or `<` in Cargo.toml.
- Use some other way in CI to ensure `serialport-rs` *can* build with old deps and MSRV rust version. Possibilities:
  - Make a dummy crate that depends on `serialport` and *does* pin the dependencies to old versions.
  - https://github.com/taiki-e/cargo-minimal-versions
  - MSRV-aware resolver https://rust-lang.github.io/rfcs/3537-msrv-resolver.html but not stable yet?

Note just bumping the MSRV is NOT enough to solve this. Bumping eases the pain a bit since it stops requiring such old dependencies, but it's still possible to run into conflicts and it'll get worse again over time as the MSRV ages. The fix is to really stop pinning versions.

sirhcel:
> Thanks [@Dirbaio](https://github.com/Dirbaio), that's exactly the point I was trying to make in [#304 (comment)](https://github.com/serialport/serialport-rs/issues/304#issuecomment-3512188873) but I feel like your wording is quite a bit more eloquent :)

For whatever reason I assumed there were only few isolated issues like with your report @Finomnis. @Dirbaio description was an eye opener for me.


> I honestly don't think anybody would complain if you released a 4.x without pins. All the users that are on an ancient Rust compiler version already need to use the Cargo.lock file to prevent breakage with a bunch of other crates, so they won't accidentally run into the problem.

[...]

PR #328, opened 2026-03-12
Unpin dependencies

sirhcel:
See #324 for details. To finally resolve all the issues from that.

I went for the `cargo update --precise` path as [cargo-minimal-versions](https://github.com/taiki-e/cargo-minimal-versions) wants to build the entire project and not just the library part with the given MSRV and we have examples requiring a higher Rust version.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,9 @@ on:
       toolchain:
         default: stable
         type: string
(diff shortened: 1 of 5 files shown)


onur-ozkan (approved):
Looks nice!

eldruin (approved):
This is great, thank you!

sirhcel:
Yes, finally! Thank you @eldruin!

PR #331, opened 2026-03-13
async support

onur-ozkan:
This adds built-in async serial support behind a new `async-io` feature. The main entry point is `SerialPortBuilder::open_async()`, which returns a cross-platform `AsyncSerialPort` implementing `AsyncRead` and `AsyncWrite`.

The async API is `futures-io` based, so the same port type can be used from different runtimes (see smol and tokio examples). The native tokio (mio based) backend will follow separately.


RossSmyth:
While an interesting idea, I wonder if we can really do a better job than the mio-serial or tokio-serial

onur-ozkan:
> While an interesting idea, I wonder if we can really do a better job than the mio-serial or tokio-serial

There is a discussion about this in the matrix channels.

FrankenApps:
> While an interesting idea, I wonder if we can really do a better job than the mio-serial or tokio-serial

Given the fact that both haven't had any commits for more than a year I would put more hope into this becoming a well-maintained alternative.

RossSmyth:
Ok, but it will require a good restructure for not-jank integration. Windows should be fairly straight-foward since it is all async internally. IDK how a "native" solution would look on Linux. epoll?

onur-ozkan:
> Ok, but it will require a good restructure for not-jank integration. Windows should be fairly straight-foward since it is all async internally. IDK how a "native" solution would look on Linux. epoll?

Just to be clear, we are not doing junk integration. It's not going to be some randomly rushed work. And yes, native solution on linux means utilizing epoll, which is already being handled in this PR through [async_io](https://docs.rs/async-io/2.6.0/async_io/). There is no reason to handle raw async primitive for every OS, we can just use async_io which is fairly stable and reliable dependency for this purpose.

onur-ozkan:
> > While an interesting idea, I wonder if we can really do a better job than the mio-serial or tokio-serial
> 
> Given the fact that both haven't had any commits for more than a year I would put more hope into this becoming a well-maintained alternative.

That was one of the main reasons we talked about before we started this implementation.

RossSmyth:
I see, well that's good.

sirhcel:
Thank you very much for getting this started @onur-ozkan! I really like the approach of re-using the already existing platform-specific serial port backends with all its arcane stuff for configuration.

When starting to give it as spin, I experienced the fallout from our dependency pinning first hand. I therefore rebased it on the current branch for unpinning all of them (#328) and bumped MSRV to 1.75.0 which we're planning to switch to with #302. See https://github.com/sirhcel/serialport-rs/tree/async-support-on-unpinned-dependencies. This includes updating your commits for updating MSRV and unpinning dependencies. If this preserves the spirit, please feel free to update this PR.

~My [first take](https://gist.github.com/sirhcel/6176995452cbca7629b3c657e099028a) on reading data with smol did not give me any data from the sender. As as these are my first steps with futures_io, could you give me a reference on how `AsyncRead` is supposed to be used with smol and tokio?~

Figured out that `smol::io::AsyncReadExt` has the read magic I was looking for and [successfully received](https://gist.github.com/sirhcel/6176995452cbca7629b3c657e099028a?permalink_comment_id=6029010#gistcomment-6029010) the first serial data. :tada: So supporting different async runtime stems from `futures_io::AsyncRead` being supported by the respective runtime. right? Like `smol::io::AsyncReadExt` extending `futures_io::AsyncRead` and `tokio_util::compat::Compat` wrapping `futures_io::AsyncRead` for their flavor?

onur-ozkan:
> Figured out that smol::io::AsyncReadExt has the read magic I was looking for and [successfully received](https://gist.github.com/sirhcel/6176995452cbca7629b3c657e099028a?permalink_comment_id=6029010#gistcomment-6029010) the first serial data. 🎉

Very cool! :)

> So supporting different async runtime stems from futures_io::AsyncRead being supported by the respective runtime. right? Like smol::io::AsyncReadExt extending futures_io::AsyncRead and tokio_util::compat::Compat wrapping futures_io::AsyncRead for their flavor?

Yes, that's exactly right.

I am planning to continue my work here later this week. I got some load from other stuff that I need to handle.

sirhcel:
We've had quite some discussion on the why of `AsyncRead::poll_close` in [the matrix chat](https://matrix.to/#/!ruSeSokIZmsariaNZc:matrix.org/$hM9N2Ni02tJtPoopRApsh86L9f8gWbQ8-rgKSqEcA_U?via=matrix.org&via=epfl.ch).

giuseppe998e:
While reviewing this PR, I noticed that the current implementation, that relies on `async-io`, presents a structural drawback for Tokio users: `async-io` creates a separate background thread for I/O operations. This approach *bypasses* Tokio’s native `mio` driver, adding an **extra** thread that introduces unnecessary overhead, which is highly inefficient (for example, in embedded systems with constrained resources).

One might consider splitting the asynchronous implementation into separate, runtime-specific features (e.g., Tokio and Smol). For example, drawing inspiration from existing work in `mio-serial`, Tokio users could utilise the native driver without the overhead of the `async-io` thread.

onur-ozkan:
> While reviewing this PR, I noticed that the current implementation, that relies on `async-io`, presents a structural drawback for Tokio users: `async-io` creates a separate background thread for I/O operations. This approach _bypasses_ Tokio’s native `mio` driver, adding an **extra** thread that introduces unnecessary overhead, which is highly inefficient (for example, in embedded systems with constrained resources).
> 
> One might consider splitting the asynchronous implementation into separate, runtime-specific features (e.g., Tokio and Smol). For example, drawing inspiration from existing work in `mio-serial`, Tokio users could utilise the native driver without the overhead of the `async-io` thread.

Could you please share the results if you have already tested or benchmarked this? For example, what exactly is the "**constrained resource**" you are referring to that would be highly inefficient with the current approach? I haven't measured anything yet, but I am fairly confident that on RP5 the difference would be on the order of microseconds. It would definitely be nice to avoid any "**highly inefficient**" overheads where possible, but I would prefer to avoid adding conditional backend complexity that could increase maintenance cost by 10x just to save a few microseconds.

Additionally, I wouldn't include this complication in the current PR. We don't need to implement everything at once. It would also be quite difficult for other maintainers to review and test massive PR. Instead, we could gate this support behind an "async_io" feature and structure the code so it can be easily extended with additional backend improvements later. If needed, we could gradually expand async support to include native tokio support as well.

It's also worth mentioning that there is already a tokio-native solution, but there is currently no way to use serialport on the smol executor. Either way, I would like to see actual numbers on real hardware before investing effort into this level of complexity.

giuseppe998e:
> Could you please share the results if you have already tested or benchmarked this? For example, what exactly is the "constrained resource" you are referring to that would be highly inefficient with the current approach? I haven't measured anything yet

I am developing on a custom, battery powered, embedded Linux (Yocto) board featuring a single-core ARMv8 processor and 256 MiB of RAM for real-time audio and AI translation (on a dedicated NPU). I'll provide benchmarks as soon as I can, but **the underlying issue is architectural**.

On a single-core system, Tokio operates most efficiently using the `current_thread` runtime, which handles task scheduling and all the system I/O (socket, serial, fs, etc.) *within a single thread*. However, because this crate utilizes `async-io` outside of its native executor, it automatically spawns an independent background thread dedicated **solely** to its own `epoll` loop.

This implementation forces a multi-threaded execution model onto a single physical core. Every I/O operation introduces **unnecessary** overhead: the OS wakes the `async-io` background thread, which processes the event and notifies the `Waker`, triggering a context switch back to the main Tokio runtime. This continuous context switching and CPU cache invalidation wastes critical clock cycles, especially since it can be avoided.

You can apply the same reasoning when using multiple threads/cores; the problem is masked, but the inefficiency remains.

example

> Additionally, I wouldn't include this complication in the current PR. [...] Instead, we could gate this support behind an "async_io" feature and structure the code so it can be easily extended with additional backend improvements later.

Put that way, I completely agree; my issue was with the initial idea of treating this as a “silver bullet” for any runtime.

I'm more than willing to add the logic myself to support Tokio natively, if needed.

onur-ozkan:
Great, I will set up the implementation so it's ready to be extended later for the tokio feature gate.

onur-ozkan:
The status quo ([copied](https://matrix.to/#/!ruSeSokIZmsariaNZc:matrix.org/$oxxG_mXJsT3j2lxPi-QcXBTppbbQbMhMgbuxlfwNQuE?via=matrix.org&via=epfl.ch) from the matrix channel):

> - I pushed the windows support too minutes ago
> - I ran the smol and tokio examples on real hardware, both worked correctly
> - I think, at this point, we should do multiple tests and start reviewing the PR in detail (specially the windows API, as I am not very good at that part)
> - Please ignore nit/tiny parts on review as there will be final polishing once we agree that it works. Try to focus on the functionality.
> - Tokio one did not show the logs as fast as smol, I wonder if that's because it went too fast and caused freezing/lagging, or it has something that hurts tokio? This particular behaviour needs deeper testing.

trappitsch:
I had some time to have a look at this PR and test drive it with `smol` and `tokio`. I found both runtimes to be working and didn't encounter any issues. Sure, these were just some basic, straightforward tests. I was mostly wondering at this point if any help would be useful in testing specific scenarios?

PR #333, opened 2026-03-13
Run Clippy for more CI targets

### LINKED ISSUE #332: Let Clippy run for more targets in CI
Clippy only checks the code which is active for the current build target. So we should run Clippy for more - if not all - build targets.

sirhcel:
Run Clippy for more CI targets and escalate warnings to CI errors (because we would miss them hidden behind a green tick mark).

* [x] Run Clippy for different targets
* [x] Run Clippy for different feature sets

This PR resolves #332.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,6 +19,7 @@ jobs:
     steps:
       - name: Lint | install dependencies
         run: |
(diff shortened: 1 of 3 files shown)


eldruin (approved):
Sounds good to me, thank you!

PR #335, opened 2026-03-22
Feature/split api 

### LINKED ISSUE #250: Reading individual USB packets.
USB data packets are usually up to 64 bytes and I would like to be able to read those packets instead of reading just raw bytes when they come (Read).

This could greatly improve building protocols for embedded firmware as you can use the same exact encoding on the device side and on the host side.

VMshall:
Resolves #250 and relates to tracking issue #302. 

While `try_clone()` allows concurrent reading and writing by duplicating the underlying OS handle, it forces developers to manage two full `SerialPort` instances, which can lead to accidental reads from the writing thread, writes from the reading thread, or conflicting configuration changes across threads. 

This PR introduces a split() method that cleanly separates a serial port into independent read and write halves, enforcing safer usage without requiring `Arc<Mutex<...>>` synchronization overhead in user space. It adds `ReadHalf<T>` and` WriteHalf<T>` wrappers in `src/lib.rs`, where `ReadHalf` is restricted to `std::io::Read` and `WriteHalf` to `std::io::Write`. The API is extended with pub `fn` split(self: Box<Self>) on `impl dyn SerialPort`, consuming the trait object to return independent halves, along with native split(self) implementations for `TTYPort` and `COMPort` to avoid heap allocation when using concrete types. 
Additionally, a `test_split` test is introduced in `tests/test_try_clone.rs` to verify correct splitting via `TTYPort::pair().split()` and ensure proper isolation of data flow between the read and write halves.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,8 @@
 //! platform-specific port object which allows access to platform-specific functionality.
 
 #![allow(clippy::uninlined_format_args)]
+// Clippy's MSRV lint can disagree with macro-expanded code from dependencies (e.g. nix ioctl helpers) even when `cargo build` at the crate MSRV still succeeds. See CI `clippy` jobs.
+#![allow(clippy::incompatible_msrv)]
 #![deny(
     clippy::dbg_macro,
     missing_docs,
@@ -913,12 +915,12 @@ impl dyn SerialPort {
     /// This function returns an error if the serial port couldn't be cloned using `try_clone`.
     pub fn split(
         self: Box<Self>,
-    ) -> Result<(ReadHalf<Box<dyn SerialPort>>, WriteHalf<Box<dyn SerialPort>>)> {
+    ) -> Result<(
+        ReadHalf<Box<dyn SerialPort>>,
+        WriteHalf<Box<dyn SerialPort>>,
+    )> {
         let cloned = self.try_clone()?;
-        Ok((
-            ReadHalf { inner: self },
-            WriteHalf { inner: cloned },
-        ))
+        Ok((ReadHalf { inner: self }, WriteHalf { inner: cloned }))
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -870,6 +870,58 @@ impl fmt::Debug for dyn SerialPort {
     }
 }
 
+/// A half of a `SerialPort` that can only be used for reading.
+///
+/// This is typically created by splitting a port into its read and write halves.
+#[derive(Debug)]
+pub struct ReadHalf<T> {
+    pub(crate) inner: T,
+}
+
+impl<T: io::Read> io::Read for ReadHalf<T> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.inner.read(buf)
+    }
+}
+
+/// A half of a `SerialPort` that can only be used for writing.
+///
+/// This is typically created by splitting a port into its read and write halves.
+#[derive(Debug)]
+pub struct WriteHalf<T> {
+    pub(crate) inner: T,
+}
+
+impl<T: io::Write> io::Write for WriteHalf<T> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.inner.write(buf)
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.inner.flush()
+    }
+}
+
+impl dyn SerialPort {
+    /// Attempts to split the `SerialPort` into independent read and write halves.
+    ///
+    /// This allows you to write and read simultaneously from the same serial
+    /// connection on different threads without synchronization overhead.
+    ///
+    /// # Errors
+    ///
+    /// This function returns an error if the serial port couldn't be cloned using `try_clone`.
+    pub fn split(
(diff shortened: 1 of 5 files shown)


sirhcel:
Thank you for starting this PR! A first glance looks like things are going in the right direction and i will have a more in-dept look in the next days.

Rebased the essence of the initial PR onto the current main branch to get rid of the noise from unrelated and already integrated changes.

sirhcel:
Have you looked into other approaches of giving each half its own port @VMshall? I'm asking because i'm wondering whether the `split()` method could be infallible? And I've got initial thoughts on whether we can safely split reading and writing at all.

Some initial thoughts on that:
* Somehow reference the original serial port
* This could mean moving it from a potential stack location to heap (`Box`, `Arc`, ...) for being able to share it without the need for carrying lifetimes along
* Maybe duplicate just the underlying file descriptor
* Add some internal low-level abstractions for reading/writing from a file descriptor and use them for reading/writing from the original port instances and from the halves
* Add support for joining the halves back, for example for changing port settings
* Maybe add support for safely changing settings from the halves as well?

I did some superficial research on what is required for safely accessing a serial port from different threads for reading and writing (at first on Linux and Windows). The first results were like "well, maybe, it depends". So we should do some more research on whether we can actually safely share the port for reading and writing. I'm fine if this is only possible for a subset of targets and we provide splitting only conditionally.

Support for joining halves back into the original port is a goal from my side because it looks somewhat awkward to me when mandating the user to close both halves and reopen the port just for changing settings.

Please don't get me wrong with all that questions and ideas. I'd like to start the discussion.

VMshall:
Hi @sirhcel , thanks for the feedback! Your questions are guaranteed are spot on.

Here’s a quick rundown of what I found:

The Windows Concurrency Catch:
On Unix, concurrent reads and writes work as expected. However, on Windows, `serialport-rs` opens COM ports synchronously. In this mode, the OS does not guarantee concurrent execution of operations on the underlying handle, and in practice they often behave as serialized for COM ports. A blocking `read` can stall a concurrent `wri`te. So even with `try_clone()`, `split()` would not provide true parallel I/O on Windows.

Infallible `split()` & `unsplit()`:
We can make `split()` infallible while avoiding `Mutex` overhead by wrapping the port in a cheap `Arc` and implementing `io::Read `/ `io::Write` for `&COMPort / &TTYPort`. This avoids handle duplication entirely and makes `unsplit(read, write)` straightforward via `Arc::try_unwrap()`, allowing users to regain exclusive access for configuration changes. This approach assumes the underlying port types do not rely on unsynchronized internal mutable state.

Given that Windows may serialize synchronous operations anyway, are we comfortable exposing a `split()` API defined as providing independent read/write handles (but not guaranteed parallelism), perhaps with clear documentation about platform specific behavior?

sirhcel:
> The Windows Concurrency Catch: On Unix, concurrent reads and writes work as expected.

That's what I'm also assuming here. Did you came across a reference stating that this is actually safe to do?


> However, on Windows, `serialport-rs` opens COM ports synchronously. In this mode, the OS does not guarantee concurrent execution of operations on the underlying handle, and in practice they often behave as serialized for COM ports. A blocking `read` can stall a concurrent `wri`te. So even with `try_clone()`, `split()` would not provide true parallel I/O on Windows.

Do you mean opening the port synchronously like in 'not using overlapped I/O'? At a first glance, it looks to me that a split port would technically make not much sense in this case as all I/O operations will be serialized by Windows.

There was some work on switching to overlapped I/O on Windows with #55. There was quite some discussion going on there and i bet we should catch up with that first and see whether overlapped I/O actually makes sense and is feasible for us.


> Infallible `split()` & `unsplit()`: We can make `split()` infallible while avoiding `Mutex` overhead by wrapping the port in a cheap `Arc` and implementing `io::Read `/ `io::Write` for `&COMPort / &TTYPort`. This avoids handle duplication entirely and makes `unsplit(read, write)` straightforward via `Arc::try_unwrap()`, allowing users to regain exclusive access for configuration changes. This approach assumes the underlying port types do not rely on unsynchronized internal mutable state.

The port implementation do not rely on mutable state but reading and writing are operations on mutable instances/borrows. We have to get this right.

I have to admit that I had not looked into how the standard library adresses this. I've seen a split operation in no_std code several times and assumed that things will be the same on std. Today I learned that this is not the case and for example `File` and `TcpStream` provide `try_clone` here. So from the operating system side we would end up with distinct file descriptors/handles for a port which can be used for reading/writing from different threads. To me this looks like reading/writing concurrently on the duplicated file descriptor/handle pair is fine.

If popular abstractions like `File` and `TcpStream` provide `try_clone` for that purpose I would like to follow that path as well. If all the interaction beyond reading and writing (like for example configuring baud rate, flushing buffers, ...) is also safe for duplicated handles, then we would not need a restricted interface for a split port and can just get away with having `try_clone`.

Would you like to do some initial research on that and we continue the discussion on that basis?


> Given that Windows may serialize synchronous operations anyway, are we comfortable exposing a `split()` API defined as providing independent read/write handles (but not guaranteed parallelism), perhaps with clear documentation about platform specific behavior?

I don't see the benefit of providing split reading/writing on Windows when these operations are serialized. Do you see some? If there is no benefit I would not provide support for splitting the port on Windows.

PR #338, opened 2026-03-25
Fix write timeout being misreported as EOF on Windows

madmo:
When WriteFile succeeds but writes 0 bytes (i.e. a timeout), the write implementation previously returned Ok(0). This caused write_all to interpret the timeout as an unexpected EOF (Error::WRITE_ALL_EOF) instead of a timeout error.

Now, when WriteFile reports success with 0 bytes written, write returns Err(TimedOut) so that callers receive the correct error.

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -248,7 +248,16 @@ impl io::Write for COMPort {
             )
         } {
             0 => Err(io::Error::last_os_error()),
-            _ => Ok(len as usize),
+            _ => {
+                if len != 0 {
+                    Ok(len as usize)
+                } else {
+                    Err(io::Error::new(
+                        io::ErrorKind::TimedOut,
+                        "Operation timed out",
+                    ))
+                }
+            }
         }
     }
 
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -208,21 +208,25 @@ impl IntoRawHandle for COMPort {
 
 impl io::Read for COMPort {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let mut len: u32 = 0;
+        let to_read: u32 = buf
+            .len()
+            .try_into()
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidInput, e))?;
+        let mut read: u32 = 0;
 
         match unsafe {
             ReadFile(
                 self.handle,
                 buf.as_mut_ptr(),
-                buf.len() as u32,
-                &mut len,
+                to_read,
+                &mut read,
                 ptr::null_mut(),
             )
         } {
             0 => Err(io::Error::last_os_error()),
             _ => {
-                if len != 0 {
-                    Ok(len as usize)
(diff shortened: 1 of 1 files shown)


sirhcel:
Thank you @madmo! Good catch!

I'm wondering whether a timeout is the only source of a successful read of zero bytes? Do you know of others? At least cancelling an I/O operation will result in `ReadFile` to return `FALSE` and we will take the error path in that case.

madmo:
> Thank you @madmo! Good catch!
> 
> I'm wondering whether a timeout is the only source of a successful read of zero bytes? Do you know of others? At least cancelling an I/O operation will result in `ReadFile` to return `FALSE` and we will take the error path in that case.

I'm not aware of any other possible case where ReadFile/WriteFile return with success but zero bytes read/written.

It is not especially mentioned in the [official docs](https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-commtimeouts?redirectedfrom=MSDN#remarks) but that is the special case of timeout with 0 bytes available to read/write.

sirhcel:
Thanks for the heads-up @madmo! Documentation is underwhelming when it comes to "fringe cases" like serial ports. After sleeping over it and another round of research I still see reading and writing zero bytes a legit case on Windows and we should keep on returning `Ok(0)` in this case.

madmo:
> Thanks for the heads-up @madmo! Documentation is underwhelming when it comes to "fringe cases" like serial ports. After sleeping over it and another round of research I still see reading and writing zero bytes a legit case on Windows and we should keep on returning `Ok(0)` in this case.

If a Ok(0) can be anything but a timeout we should probably also think about/remove the special handling in the io::Read implementation of COMPort.

But I have to admit that I'm not a Windows expert and definitely not a Windows COM port expert. This change works for my use case, and I'm absolutely fine not moving this change upstream (You can close this PR if you want to keep things tidy).

sirhcel:
> If a Ok(0) can be anything but a timeout we should probably also think about/remove the special handling in the io::Read implementation of COMPort.

We absolutely should change the implementation of `io::Read` as well. I expect a "symmetrical" behavior from both. I also did not notice that there was already a check for `len != 0` in the first place.


> But I have to admit that I'm not a Windows expert and definitely not a Windows COM port expert. This change works for my use case, and I'm absolutely fine not moving this change upstream (You can close this PR if you want to keep things tidy).

Me neither. That's why I'm doing some research and asking a lot of questions in this case. 😅 Your change looks reasonable and I'd like to integrate it. Things are not tidy at all here. But I'd like to clean up as we go.

So I would like to go in the direction of https://github.com/sirhcel/serialport-rs/commit/f95cdc05fa4a18b44e79318b2fd48185b2665715. I did not manage to propose this as a suggestion in a review for this PR. Please feel free to cherry-pick or I can push it to this PR here as well.

madmo:
I cherry-picked your suggested change and updated the PR. I have to admit that I completely missed the case for intentional zero byte writes/reads!

Thanks a lot for the support!

PR #339, opened 2026-03-25
feat: add port_info() default method to SerialPort trait

### LINKED ISSUE #276: Allow getting `SerialPortInfo` from `dyn SerialPort`
At the moment there is no way to get the info of a serial port other than enumerating all of them. Would it be possible to add the ability to get the info from `dyn SerialPort` directly?

KaustubhOG:
Closes #276

## Problem

There is currently no way to retrieve `SerialPortInfo` from an open`dyn SerialPort` without calling `available_ports()` manually and searching by name.

## Solution

Add a default `port_info()` method to the `SerialPort` trait. The implementation calls `self.name()`, enumerates available ports, and returns the matching `SerialPortInfo` entry.

Since it is a default method, all existing implementors (`TTYPort`,`COMPort`, and the `&mut T` blanket impl) get it for free with no changes required on their end.

Returns `Err` if:
- the port has no name (e.g. virtual/pseudo-terminal ports)
- `available_ports()` fails
- no matching port is found

## Notes

- All existing tests pass
- Platform-specific types can override this method in the future
  if a faster implementation is needed (e.g. caching info at open time)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -472,6 +472,22 @@ pub trait SerialPort: Send + io::Read + io::Write {
     /// Additionally it may not exist for virtual ports.
     fn name(&self) -> Option<String>;
 
+    /// Returns the [`SerialPortInfo`] for this port by matching its name against
+    /// the results of [`available_ports()`].
+    ///
+    /// This may be slow on some platforms as it enumerates all available ports.
+    /// Returns an error if the port has no name, if enumeration fails, or if the
+    /// port is not found among the available ports (e.g. virtual or pseudo-terminal ports).
+    fn port_info(&self) -> Result<SerialPortInfo> {
+        let name = self
+            .name()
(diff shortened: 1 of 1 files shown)


KaustubhOG:
Should this target the playground-5.0 branch instead of main, given that it touches the SerialPort trait? i read some old pr where you listed it specific point

sirhcel:
Thank you for starting this draft PR @KaustubhOG! Please see my comment https://github.com/serialport/serialport-rs/issues/276#issuecomment-4151023913 in the original issue.

Regarding this PR: Enumerating all ports for identifying the details of a single one looks somewhat like a brute force approach. Sure, this is what we could do with just the current public API. Did you have a look into how information for a certain port could be fetched with the enumeration mechanisms we are using (like libudev, macOS' I/O registry, `SetupDiEnumDeviceInfo` on Windows, ...)? I would favor such an approach because it would generate only a fraction of the runtime cost of enumerating all serial devices.

PR #322, opened 2026-03-05
Unhandled return values in TTYPort::open

NONnonHere:
@sirhcel Can you check this 
Is it solving the TTYPort issue raised.

Issue #321, opened 2026-03-05
Unhandled return values in TTYPort::open


sirhcel:
While reviewing #316, I noticed several unhandled results from FFI calls in `TTYPort::open`. For example [here](https://github.com/serialport/serialport-rs/blob/98e5eeb8698756e6af2055848b8d423205389338/src/posix/tty.rs#L154C9-L154C15) when passing termios configuration down to the OS. And there are more ...

We should clean this up, and while at it pay attention to closing the file descriptor if needed. It seems that we currently have a mix of explicitly closing it and relying on our own `OwnedFd` to ensure that.

sirhcel:
Resolved by #322 and released with [4.9.0](https://crates.io/crates/serialport/4.9.0).

PR #344, opened 2026-04-11
Prevent HANDLE leaks in COMPort::open using OwnedHandle

NONnonHere:
This PR implements the Windows equivalent of the POSIX file descriptor fix merged in #322 (and discussed in #321). 

Previously, if `COMPort::open` successfully called `CreateFileW` but failed during subsequent configuration steps (like `SetupComm` or `get_dcb`), it would return early and leak the raw `HANDLE`.

**Changes:**
* Upgraded `COMPort` to store an `OwnedHandle` instead of a raw `HANDLE`.
* Immediately wrapped the handle generated by `CreateFileW` into an `OwnedHandle` so that any early returns automatically trigger `CloseHandle` via RAII.
* Updated all WinAPI FFI calls to use `.as_raw_handle()`.
* Removed the manual `Drop` implementation since `OwnedHandle` manages the lifecycle natively.
* Fixed `try_clone_native` to safely wrap the duplicated handle.

*Note* Local `cargo clippy` throws `incompatible_msrv` warnings because `OwnedHandle` stabilized in 1.63.0.

### DIFF
--- a/src/windows/com.rs
+++ b/src/windows/com.rs
@@ -1,5 +1,7 @@
 use std::mem::MaybeUninit;
-use std::os::windows::prelude::*;
+use std::os::windows::io::{
+    AsHandle, AsRawHandle, BorrowedHandle, FromRawHandle, IntoRawHandle, OwnedHandle, RawHandle,
+};
 use std::time::Duration;
 use std::{io, ptr};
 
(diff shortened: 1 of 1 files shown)


RossSmyth:
Seems like a fine change.

sirhcel:
Good catch! Thank you very much!

As we need an MSRV bump for `OwnedHandle`, could you please change the target branch to playground-5.0 @NONnonHere?

src/windows/com.rs
```diff
@@ -154,26 +161,20 @@ impl COMPort {
...
-        match unsafe { GetCommModemStatus(self.raw_handle(), &mut status) } {
+        match unsafe { GetCommModemStatus(self.handle.as_raw_handle(), &mut status) } {
             0 => Err(super::error::last_os_error()),
             _ => Ok(status & pin != 0),
         }
     }
 
-    fn open_from_raw_handle(handle: RawHandle) -> Self {
-        // It is not trivial to get the file path corresponding to a handle.
-        // We'll punt and set it `None` here.
```
sirhcel, line 165: Could you please leave this comment in place? This one is really useful in the sense that it explains the why.

src/windows/com.rs
```diff
@@ -1,5 +1,7 @@
 use std::mem::MaybeUninit;
-use std::os::windows::prelude::*;
```
sirhcel, line 2: I really like making imports explicit.

src/windows/com.rs
```diff
@@ -439,23 +452,23 @@ impl COMPort {
...
 
-        if unsafe { PurgeComm(self.raw_handle(), buffer_flags) != 0 } {
+        if unsafe { PurgeComm(self.handle.as_raw_handle(), buffer_flags) != 0 } {
             Ok(())
         } else {
             Err(super::error::last_os_error())
         }
     }
 
-    pub fn set_break(&self) -> Result<()> {
```
sirhcel, line 449: This seems unrelated to the topic of this PR to me. What's the reason behind this change? In case it is a follow-up to #352, please put it in a separate PR.

src/windows/com.rs
```diff
@@ -439,23 +452,23 @@ impl COMPort {
...
+        if unsafe { SetCommBreak(self.handle.as_raw_handle()) != 0 } {
             Ok(())
         } else {
             Err(super::error::last_os_error())
         }
     }
 
-    pub fn clear_break(&self) -> Result<()> {
-        if unsafe { ClearCommBreak(self.raw_handle()) != 0 } {
+    pub(crate) fn clear_break(&self) -> Result<()> {
```
sirhcel: Same as before.

sirhcel (changes requested):
Thank you very much for rebasing the PR! I've still got some nitpicking, could you please have a look at it?

sirhcel:
Thank you very much!

PR #349, opened 2026-05-06
Preserve raw Windows OS error codes

newbeeeeeeee-cpu:
## Summary

This change preserves the raw OS error code on `serialport::Error` and keeps it available across Windows error wrapping paths.

## What changed

- add `raw_os_error: Option<i32>` to `serialport::Error`
- add `Error::raw_os_error()` as a public accessor
- preserve `io::Error::raw_os_error()` in `From<io::Error> for Error`
- preserve the raw OS error again when converting `serialport::Error` back into `std::io::Error`
- update the Windows `last_os_error()` wrapper to store the Win32 error code and include `(os error N)` in the human-readable description
- add tests covering:
  - default errors without OS codes
  - conversion from `io::Error`
  - conversion back into `io::Error`
  - Windows wrapper behavior

## Why this change

The important part of a Windows serial failure is the original OS error code, not the localized error message text.

Today the crate maps Windows failures into a localized description string, but drops the original numeric error code. That makes it harder for callers to reliably distinguish cases like device busy or access denied across machines with different Windows display languages.

By preserving the raw OS error code on `serialport::Error`, callers can match on a stable value that is not affected by Windows internationalization. The localized description is still useful for display, but the error code is the durable identifier for diagnostics, logging, and application-level handling.

## Behavior notes

This keeps the existing `ErrorKind` mapping in place, but also exposes the original OS error code to callers that need more precision.

On Windows, error strings now include the standard `(os error N)` suffix so the code is visible even when only the formatted message is logged.

## Testing

- `cargo test`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,43 +88,20 @@ pub struct Error {
     pub kind: ErrorKind,
     /// A description of the error suitable for end-users
     pub description: String,
-    /// The raw operating system error code, if one was provided by the OS.
-    ///
-    /// This is useful when callers need a stable identifier for the failure that
-    /// does not depend on the current system language.
-    pub raw_os_error: Option<i32>,
 }
 
 impl Error {
     /// Instantiates a new error
     pub fn new<T: Into<String>>(kind: ErrorKind, description: T) -> Self {
-        Self::new_with_os_error(kind, description, None)
-    }
-
-    /// Instantiates a new error with an optional raw operating system error code.
-    pub fn new_with_os_error<T: Into<String>>(
-        kind: ErrorKind,
-        description: T,
-        raw_os_error: Option<i32>,
-    ) -> Self {
         Error {
             kind,
             description: description.into(),
-            raw_os_error,
         }
     }
 
     /// Returns the corresponding `ErrorKind` for this error.
     pub fn kind(&self) -> ErrorKind {
         self.kind
-    }
-
-    /// Returns the raw operating system error code, if one is available.
-    ///
-    /// This is useful when callers need a stable identifier for the failure that
-    /// does not depend on the current system language.
-    pub fn raw_os_error(&self) -> Option<i32> {
-        self.raw_os_error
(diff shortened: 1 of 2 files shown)


sirhcel:
Thank you for opening this PR! I see your point.

For whatever reason there was no CI run in the first place. Closing and reopening finally triggered the CI pipeline. Looking at the pipeline run, this is a semver-breaking change which we can't integrate into the current 4.x releases as-is.

For leaving the public interface of `serialport::Error` alone, I would have tried providing the raw error for Windows as an extension trait for it. But i was surprised to discover that `serialport::Error` is generic for all backends and has all public fields which does not play well with adding an extension trait right on the spot. Let me sleep over that.

Another thing which puzzles me is `Error::raw_os_error` being `Option<i32>` when [`GetLastError`](https://docs.rs/windows-sys/0.52.0/windows_sys/Win32/Foundation/fn.GetLastError.html) returns [WIN32_ERROR](https://docs.rs/windows-sys/0.52.0/windows_sys/Win32/Foundation/type.WIN32_ERROR.html) aka `u32`. What is the idea behind that? I'm asking because the [Remarks section](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror#remarks) for `GetLastErrorr` explicitly states the unsigned and 32 bit data type and I see the conversion critical from this angle.

newbeeeeeeee-cpu:
Thanks for the feedback. I updated the PR to keep it 4.x-compatible: no public serialport::Error field/API changes now, only the Windows last_os_error() description gets the Win32 code appended as (os error N).

I also kept the value as the u32 returned by GetLastError(). The earlier i32 shape was only following std::io::Error::raw_os_error() / std::io::Error::from_raw_os_error(), not implying Win32 errors are signed.

Validated with:
- cargo fmt --check
- cargo test
- cargo check --target x86_64-pc-windows-gnu

PR #350, opened 2026-05-10
add port-chain feature for usb devices

### LINKED ISSUE #222: Physical port for usb serial devices

xobs:
Add a feature to list a USB port's chain. This gives a unique value to the port, and allows for differentiating between different serial ports that may share the same VID/PID/Serial.

The approach is mostly taken from the nusb project, with adjustments made to adapt it to the serialport way of doing things.

This closes #222.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,8 @@
 use std::error::Error as StdError;
 use std::fmt;
 use std::io;
+#[cfg(feature = "usbportinfo-location")]
+use std::num::{IntErrorKind, ParseIntError};
 use std::str::FromStr;
 use std::time::Duration;
 
@@ -845,22 +847,6 @@ pub struct Location {
 }
 
 #[cfg(feature = "usbportinfo-location")]
-impl core::fmt::Display for Location {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::fmt::Result {
-        // Use a location format similat to Linux' sysfs 'bus-port.port[....]'.
-        //
-        // TODO: Switch to Iterator::intersperse when it get stabilized and available to us.
-        let port_chain = self
-            .port_chain
-            .iter()
-            .map(|p| p.to_string())
-            .collect::<Vec<_>>()
-            .join(".");
-        write!(f, "{}-{}", self.bus_id, port_chain)
-    }
-}
-
-#[cfg(feature = "usbportinfo-location")]
 impl Location {
     /// Create a new USB device location based on some string identifying the bus number, along with
     /// the path taken to a particular port.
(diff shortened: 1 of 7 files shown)


xobs:
The semver checks fail because they're building with `--all-features`, however the new features are gated off which I think doesn't impact semver as it's not a default feature.

The CI checks fail because libc now has an MSRV of 1.65.0.

sirhcel:
Thank you very much for this PR @xobs! This looks good to me. I took the freedom to harmonize the feature name with the already existing addition to `UsbPortInfo`.

Also for making me look in the semver issues in CI one more time. I did not see the woods for the trees in the first place and now figured out that Clippy just got a new MSRV lint which generated these warnings. We've been running the MSRV checks with pinned dependencies in CI but not Clippy yet. I fixed this in the main branch and rebased this PR on top.

src/lib.rs
```diff
@@ -825,6 +825,12 @@ pub struct UsbPortInfo {
     pub manufacturer: Option<String>,
     /// Product name (arbitrary string)
     pub product: Option<String>,
+    /// ID of the USB bus where this device is connected
+    #[cfg(feature = "usbportinfo-chain")]
+    pub bus_id: String,
+    /// Physical port heirarchy
+    #[cfg(feature = "usbportinfo-chain")]
+    pub port_chain: Vec<u8>,
```
sirhcel: I'm wondering whether it will improve ergonomics for filtering ports when factoring out these two fields into a struct named for example`Location`. Just thinking about looking a port at a certain "location", I'm assuming that I'm always looking for bus ID and the chain together. With a dedicated struct, I could setup my filter criterion once and then just compare both at once using `==`.

Like for example

```rust
let location = Location::new("00", vec![1, 2, 3]);

if info.location == location {
    // We found it
}
```

How do you work with these fields? Would you consider such a grouping beneficial to you?

xobs:
I actually tend to use them in a struct like this:

```rust
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
struct PortAddress {
    bus_id: String,
    port_chain: Vec<u8>,
}

impl core::fmt::Display for PortAddress {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{} : {:?}", self.bus_id, self.port_chain)
    }
}

impl PortAddress {
    pub fn new(info: &nusb::DeviceInfo) -> Self {
        PortAddress {
            bus_id: info.bus_id().to_owned(),
            port_chain: info.port_chain().to_owned(),
        }
    }
    pub fn parent(&self) -> Option<PortAddress> {
        if self.port_chain.len() <= 1 {
            return None;
        }
        Some(PortAddress {
            bus_id: self.bus_id.clone(),
            port_chain: self.port_chain[0..self.port_chain.len() - 1].to_owned(),
        })
    }
}
```

...and then I sling `PortAddress` objects everywhere. You're right in that one of the fields is usually not that interesting alone.

sirhcel:
Please excuse the back and forth with my review comment. I picked the wrong spot for annotating in the first place.

Alright. This sounds like we should add a `Location` type because you already have your bespoke one and other users will likely need one as well.

Now the question is: who? Do you want to give it a try? If not, I can do it.

xobs:
I took a swing at it based on the wrapper I'm currently using.

(I also removed some stray debugging things that snuck in.)

xobs:
Let me know if there's anything else you wanted me to change, or if you had a different design in mind for `Location`.

sirhcel:
Thank you very much for grouping the fields into `Location`! This looks good to me. While playing with this branch, the following things came into my mind:

1. What about not inventing the next format for displaying the location but follow the one Linux uses in sysfs? Or do you see a good reason for hand-rolling our own display format?
2. I would harmonize the preview feature name with that from the field as it is already done with `usbportinfo-interface`
3. When implementing `Display` for Location, I would also like to implement `FromStr` 

I already played with the first two in the branch [add-port-info](https://github.com/sirhcel/serialport-rs/commits/add-port-info/). What do you think about these points?

xobs:
Those changes look good to me.

sirhcel:
Great! Finally added the `FromStr` implementation for `Location`. Do you see the need for restricting the fields of the latter to be non-empty? If not, I would would leave it as-is. This leaves some corner cases which do not please my esthetics  but allow formatting and parsing to be straight forward and I see benefit in the latter.

When writing tests, i stumbled upon `Location::is_child_of`: To my understanding, the term only refers to direct descendants and I would propose `Location::is_descendant_of` instead. Do you have another idea?

Besides this nitpicking, this PR looks finally ready for merging to me. Thank you very much @xobs!

xobs:
No, `is_descendent_of()` works for me. I could maybe see Location being empty if you don't have access to the USB filesystem on Linux, but then a whole bunch of other things like VID/PID would also disappear. So it seems fine to me.

sirhcel:
Thank you very much @xobs!

PR #354, opened 2026-06-04
Various minor cleanups

qwandor:
This mostly tidies up some things after the MSRV change, along with some formatting fixes.


sirhcel:
Thank you very much!

PR #355, opened 2026-06-04
Update to nix 0.31

qwandor:
This rebases the work in #263 on top of the playground-5.0 branch and adds some cleanups, and moves to the new nix `Flock` API.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -234,7 +234,6 @@ jobs:
   aarch64-apple-darwin:
     uses: ./.github/workflows/build.yaml
     with:
-      disable_tests: true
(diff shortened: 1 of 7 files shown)


sirhcel:
Thank you very much for working on switching to nix' new `Flock` API!

I rebased it on top of the now integrated #354. Last but not least for heaving CI checks again. I added fixes for tests which failed. And while at it I also did some minor cleanup of some older convoluted test code on the side of the road.

The PR looks now good to integrate. I will sleep over my change to `test_osx_pty_pair` before doing so.

sirhcel:
Thank you very much for your conribution @paolobarbolini and @qwandor!

Issue #346, opened 2026-04-14
Option to leave baud rate (and other settings) as-is.


ryancdotorg:
Based on the existing API, it doesn't look like there's a clean way to do this, but perhaps an API could be added to get the current settings for a serial port?

I want to be able to open a port that's already configured, inheriting some/all of the existing settings. Many `getty` type programs, for example, do this.

sirhcel:
This is a good point @ryancdotorg! I was already thinking about this from the perspective of USB devices using [CDC ACM](https://en.wikipedia.org/wiki/USB_communications_device_class#Abstract_Control_Model) for communicating with the host but not bridging a physical UART. In this setting, baud rate setting modeled by the CDC ACM protocol has no meaning and setting a dummy value should not be mandated.

This is something we should address with our next major release #302. What about refactoring the builder to make the baud rate optional and adding `SerialPort::new(path: impl Into<Cow<'a, str>>)` to just instantiate a builder with a port name but no baud rate value? We could also leave `serialport::new` and deprecate it with a hint to the former.

PR #359, opened 2026-07-20
propagate tcgetattr errors when reading baud rate

NONnonHere:


sirhcel:
Thank you for getting this started. I would propagate errors from not being able to determine the baud rate.

I was not expecting that the `SerialPort` trait from the main branch already returns a `Result<u32>` from `SerialPort::baud_rate` and now I'm puzzled why we store tha last set baud rate on Apple targets at all. Please excuse my back-and-forth rebasing caused by my initial assumption.

src/posix/tty.rs
```diff
@@ -441,8 +451,11 @@ impl FromRawFd for TTYPort {
         // It's not guaranteed that the baud rate in the `termios` struct is correct, as
         // setting an arbitrary baud rate via the `iossiospeed` ioctl overrides that value,
         // but extract that value anyways as a best-guess of the actual baud rate.
+        //
+        // FromRawFd::from_raw_fd can not fail, so fall back to reporting a baud rate of 0  when it
+        // can not be determined.
         #[cfg(any(target_os = "ios", target_os = "macos"))]
-        let baud_rate = get_termios_speed(fd.as_raw_fd());
+        let baud_rate = get_termios_speed(fd.as_raw_fd()).unwrap_or(0);
```
sirhcel: Using in-band signaling feels a bit sketchy to me. Not having information about the baud rate is fine, but I would propagate this as an error to the user and not making it looking in-band for magic numbers.

The `baud_rate` field is used on Apple targets for tracking the value initially configured. I'm seeing the two things here:

1. There has been already discussion about leaving the baud rate alone when opening a serial port (#346, and I think we had this discussion already way back in the past but I don't find it right now). Shouldn't we attempt to read back the actual baud rate in this case?

2. Couldn't we get rid of the `baud_rate` field at all? I we are using `get_termios_speed` in one place, wouldn't it be the right thing to consistently use it everywhere? In the latter case we could get rid of the field and propagate errors from determining the baud rate as it happens.

There is no counterpart for the BSD/Apple ioclt `IOSSIOSPEED`. But it looks to me that we can read back the actual baud rate from the `termios` struct's field on Darwin as these fields carry the baud rate as an integer value and not as an encoded enum.

sirhcel: This references code from my back-and-forth rebasing. But the questions stands for an integration into main as well.

sirhcel:
Thank you for updating the PR! The hardware tests run just fine on my machine with SiLabs CP2102N devices.

I did some research on `IOSSIOSPEED` and `tcsetattr`/`tcgetattr` stumbled upon the following pieces of information.

1. Apple's [`SerialPortSample_SerialPortSample_c.html`](https://developer.apple.com/library/archive/samplecode/SerialPortSample/Listings/SerialPortSample_SerialPortSample_c.html) states that a baud rate set via the `IOSSIOSPEED` ioctl won't be returned by `cfgetispeed` and `cfgetospeed`. But to my understanding the latter two functions return only the value previously read into a `struct termios` via `tcgetattr` which is called before the `IOSSIOSPEED` ioctl and therefore the ioctl won't change these values at all. Might it be the case that this example draws the wrong conclusion stated below?

    > ```c
    > // Print the new input and output baud rates. Note that the IOSSIOSPEED ioctl interacts with the serial driver
    > // directly bypassing the termios struct. This means that the following two calls will not be able to read
    > // the current baud rate if the IOSSIOSPEED ioctl was used but will instead return the speed set by the last call
    > // to cfsetspeed.
    > ```

2. Apple's test [`_testIOSSIOSPEEDIoctl`](https://github.com/apple-oss-distributions/IOSerialFamily/blob/1c12ee3d9ec665bb53d0151644408d25e54fe012/tests/IOSerialTestLib.c#L31) from `IOSerialTestLib.c` assumes that it can read back a baud rate set via the `IOSSIOSPEED` ioctl via `tcgetattr` plus `cfgetispeed`/`cfgetospeed`. There is no mention of driver-specific behavior and the test case checks that this round trip returns the very same baud rate.

3. Our documentation from [`NOTES.md` for macOS / iOS](https://github.com/serialport/serialport-rs/blob/53ac54b42a941d14504b609330dd62b11a3d01ea/NOTES.md#macos--ios) states that the round trip `tcgetattr` followed by `tcsetattr` is not possible after setting a custom baud rate via `IOSSIOSPEED`. But it does not state anything on `IOSSIOSPEED` followed by `tcgetattr`.

4. Our tests from [`tests/test_baudrate.rs`](https://github.com/serialport/serialport-rs/blob/53ac54b42a941d14504b609330dd62b11a3d01ea/tests/test_baudrate.rs#L24) which read back the actual baud rate run just fine with reading back the baud rate via `tcgetattr`. At least for the baud rates in the test and with SiLabs CP2102N USB serial adapters.

At a first glance, dropping the internal field and reading the baud rate directly via `tcgetattr` seems fine. At least in my setup. So what conclusion to draw from that?

NONnonHere:
> Thank you for updating the PR! The hardware tests run just fine on my machine with SiLabs CP2102N devices.
> 
> I did some research on `IOSSIOSPEED` and `tcsetattr`/`tcgetattr` stumbled upon the following pieces of information.
> 
> 1. Apple's [`SerialPortSample_SerialPortSample_c.html`](https://developer.apple.com/library/archive/samplecode/SerialPortSample/Listings/SerialPortSample_SerialPortSample_c.html) states that a baud rate set via the `IOSSIOSPEED` ioctl won't be returned by `cfgetispeed` and `cfgetospeed`. But to my understanding the latter two functions return only the value previously read into a `struct termios` via `tcgetattr` which is called before the `IOSSIOSPEED` ioctl and therefore the ioctl won't change these values at all. Might it be the case that this example draws the wrong conclusion stated below?
>    > ```c
>    > // Print the new input and output baud rates. Note that the IOSSIOSPEED ioctl interacts with the serial driver
>    > // directly bypassing the termios struct. This means that the following two calls will not be able to read
>    > // the current baud rate if the IOSSIOSPEED ioctl was used but will instead return the speed set by the last call
>    > // to cfsetspeed.
>    > ```


Agreed on (1): the sample looks at an old `options`, therefore cannot say anything about the driver behavior. Its own comment "will instead return the speed set by the last call to `cfsetspeed`" is merely describing the situation

> 2. Apple's test [`_testIOSSIOSPEEDIoctl`](https://github.com/apple-oss-distributions/IOSerialFamily/blob/1c12ee3d9ec665bb53d0151644408d25e54fe012/tests/IOSerialTestLib.c#L31) from `IOSerialTestLib.c` assumes that it can read back a baud rate set via the `IOSSIOSPEED` ioctl via `tcgetattr` plus `cfgetispeed`/`cfgetospeed`. There is no mention of driver-specific behavior and the test case checks that this round trip returns the very same baud rate.

(2) is more than “no mention of driver-specific behavior” `_testIOSSIOSPEEDIoctl` checks that `cfgetispeed(&options) == speed && cfgetospeed(&options) == speed` strictly, for non-standard speeds 10400, 31250, 40000....... 
(100000→115200 is 15%)
Exact equality on non-standard values rules out rounding, not just staleness


> 3. Our documentation from [`NOTES.md` for macOS / iOS](https://github.com/serialport/serialport-rs/blob/53ac54b42a941d14504b609330dd62b11a3d01ea/NOTES.md#macos--ios) states that the round trip `tcgetattr` followed by `tcsetattr` is not possible after setting a custom baud rate via `IOSSIOSPEED`. But it does not state anything on `IOSSIOSPEED` followed by `tcgetattr`.


(3) supports, instead of being silent: The problem NOTES.md mentions is switching from `tcgetattr` to `tcsetattr` round trip. That can only go wrong because the struct it returns contains the non standard speed that `tcsetattr` doesn’t accept. A standard speed, rounded, would be fine so the note already assumes `tcgetattr` gives the right speed.


> 4. Our tests from [`tests/test_baudrate.rs`](https://github.com/serialport/serialport-rs/blob/53ac54b42a941d14504b609330dd62b11a3d01ea/tests/test_baudrate.rs#L24) which read back the actual baud rate run just fine with reading back the baud rate via `tcgetattr`. At least for the baud rates in the test and with SiLabs CP2102N USB serial adapters.
> 


(4) with the catch that it does not affect the result. The `check_baud_rate` function tolerates ±0.5% thus
it is not really a exact check. However, the non-standard rates will fail loudly 

> At a first glance, dropping the internal field and reading the baud rate directly via `tcgetattr` seems fine. At least in my setup. So what conclusion to draw from that?









So dropping the field is sound and it matches what NOTES.md already states as the design the `termios` struct isn't cached, the kernel is the canonical source, The field was the last piece of exactly that cached state

sirhcel:
After sleeping over it: What about creating a branch for testing the removal of the Apple-only baud rate cache field? For what we've seen here so far, I bet we are good with removing this field and reading the baud rate via `tcgetattr` and `cfgetXspeed`. But I would like to ask the community for a test drive. I have only a handful of USB serial adapters and a single version of macOS at hand. May be there is some more insight waiting for us in the field.

May be we should add an example binary for this task so that people who are willing to help out just need to clone and run a single example.

PR #360, opened 2026-07-25
Make UsbPortInfo::location optional

sirhcel:
I overlooked the use of default values where an `Option<Location>` would be more idiomatic in #350. I also overlooked some issues with MSRV builds. This PR cleans these things up.

Semver checks are expected to fail because they are not taking into account that `UsbPortInfo::location` is preview feature-gated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,13 +98,13 @@ project adheres to [Semantic Versioning](https://semver.org/).
   [#251](https://github.com/serialport/serialport-rs/issues/251)
   [#243](https://github.com/serialport/serialport-rs/issues/243)
   [#239](https://github.com/serialport/serialport-rs/pull/239)
-  [#29](https://github.com/serialport/serialport-rs/pull/29)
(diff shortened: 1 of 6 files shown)


sirhcel:
@xobs I missed some things when landing #350 and cleaned them up here. What's your opinion on that?

I've noticed along the way that the bus IDs are rendered with leading zeroes of different length on Linux and macOS/Apple. Is there a reason behind that? If not, I would prefer to render them without leading zeroes.

xobs:
Seems fine to me. I think the numbers are just how they're usually printed on that platform. Either way works equally well!

sirhcel:
Thank you very much for the quick reply! Let's keep formatting bus IDs with leading zeroes then.

sirhcel:
Sorry @xobs, I messed up my edits. Formatting bus IDs without leading zeroes already landed in in the main branch and after looking into the actual string representations from libudev (with leading zeroes) and sysfs (without leading zeroes) disagree at least on my machine, I just fixed the inconsistent state by formatting without leading zeroes.

PR #365, opened 2026-07-27
Make platform-specific serial parameters conditional at compile time

NONnonHere:
Implements the approach discussed in #302 for PRs #189 and #195 instead of exposing unsupported variants and failing at runtime, gate them behind `#[cfg()]` and mark the enums `#[non_exhaustive]`.

## Changes
 
**Parity** (from #189, original work by @YgorSouza)
- `Parity::Mark` and `Parity::Space` gated to `windows`, `linux`, `android`
- `Parity` marked `#[non_exhaustive]`
- POSIX: clears/sets `CMSPAR` flag on linux/android
- Windows: maps to `MARKPARITY`/`SPACEPARITY`

**StopBits** (from #195, original work by @robdobsn)
- `StopBits::OnePointFive` gated to `windows`
- `StopBits` marked `#[non_exhaustive]`
- Windows: maps to `ONE5STOPBITS`

## Downstream impact analysis
 
Scanned 300+ dependent crates of serialport 4.9.0 for usage patterns, weighted by crates.io download counts.
 
**Parity usage (134 crates reference it):**
- ~45 only set parity via the builder (write-only, unaffected)
- ~11 read parity back via `.parity()`
- ~29 match on parity, but only ~9 match on `serialport::Parity` directly the rest match on their own wrapper types
- 1 crate uses Mark/Space (`rtcom-core`, 143 downloads) collapses both to `Parity::None` silently

**StopBits:** 168 crates reference `StopBits::One`, zero reference `OnePointFive` (doesn't exist yet).

**Tested against downstream crates** by patching their `Cargo.toml` to point at this branch:
- `stuart-cli`: `E0004: non-exhaustive patterns` 
- `rfc2217-rs`: `E0004: non-exhaustive patterns` 
- `serialport-stream`: `E0004: non-exhaustive patterns` 
All got compile-time errors forcing explicit handling of new variants no silent runtime failures.

### DIFF
--- a/examples/hardware_check.rs
+++ b/examples/hardware_check.rs
@@ -227,6 +227,7 @@ fn test_single_port(port: &mut serialport::SerialPort, loopback: bool) {
     // Test setting stop bits
     println!("Testing stop bits...");
     stop_bits_check!(port, StopBits::Two);
+    stop_bits_check!(port, StopBits::OnePointFive);
     stop_bits_check!(port, StopBits::One);
 
     // Test bytes to read and write
--- a/examples/hardware_check.rs
+++ b/examples/hardware_check.rs
@@ -227,6 +227,7 @@ fn test_single_port(port: &mut serialport::SerialPort, loopback: bool) {
     // Test setting stop bits
     println!("Testing stop bits...");
     stop_bits_check!(port, StopBits::Two);
+    #[cfg(target_os = "windows")]
     stop_bits_check!(port, StopBits::OnePointFive);
     stop_bits_check!(port, StopBits::One);
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,6 +211,16 @@ pub enum Parity {
 
     /// Parity bit sets even number of 1 bits.
     Even,
+
+    /// Parity bit is set to 1.
+    ///
+    /// Only supported on Windows and Linux.
+    Mark,
+
+    /// Parity bit is set to 0.
+    ///
+    /// Only supported on Windows and Linux.
+    Space,
 }
 
 impl fmt::Display for Parity {
@@ -219,6 +229,8 @@ impl fmt::Display for Parity {
             Parity::None => write!(f, "None"),
             Parity::Odd => write!(f, "Odd"),
             Parity::Even => write!(f, "Even"),
+            Parity::Mark => write!(f, "Mark"),
+            Parity::Space => write!(f, "Space"),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -246,11 +246,13 @@ impl fmt::Display for Parity {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[non_exhaustive]
 pub enum StopBits {
     /// One stop bit.
     One,
 
     /// One and a half stop bits.
+    #[cfg(target_os = "windows")]
     OnePointFive,
 
     /// Two stop bits.
@@ -261,6 +263,7 @@ impl fmt::Display for StopBits {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match *self {
             StopBits::One => write!(f, "One"),
+            #[cfg(target_os = "windows")]
             StopBits::OnePointFive => write!(f, "OnePointFive"),
             StopBits::Two => write!(f, "Two"),
         }
@@ -271,6 +274,7 @@ impl From<StopBits> for u8 {
     fn from(value: StopBits) -> Self {
         match value {
             StopBits::One => 1,
+            #[cfg(target_os = "windows")]
             StopBits::OnePointFive => 1,
             StopBits::Two => 2,
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -250,6 +250,9 @@ pub enum StopBits {
     /// One stop bit.
     One,
 
+    /// One and a half stop bits.
+    OnePointFive,
+
     /// Two stop bits.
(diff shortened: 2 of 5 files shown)


Skgland:
> Implements the approach discussed in #302 

Are you referring to this bullet point? Or is there a comment I am missing?

> - Can we add this in a manner where we are going to expose parity modes like mark and space only on platforms actually supporting it

I find conditional enum variants problematic:

- One might accidentally not handle that variant when developing on a system where that variant is disabled. The error then only gets caught downstream (or in CI) not at compile time **on the developers machine**.
- Every match arm matching a conditional variant now needs to repeat the conditional. 

One alternative would be to make those variants not constructible but still nameable when unsupported.
E.g. by making it a `#[non_exhaustive]` struct/tuple and adding a field of a valueless type when unsupported. Though this requires adding a constructor function for supported targets.

Playground example for `StopBits` enum:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d33554945a7d86532d6a20feeeba21c6

Skgland:
> Every match arm matching a conditional variant now needs to repeat the conditional.

|  | match arm cfg(false) | match arm cfg(true) |
|--------|--------|--------|
| enum variant cfg(false) | 🟰 🆗 | compilation error |
| enum variant cfg(true) | fall through to wildcard arm (required due to `#[non_exhausive]`) | 🟰 🆗  |

sirhcel:
Thank you for preparing this PR @NONnonHere! And thank you for your feedback and pointing out an interresting potential alternative @Skgland! I see your point and Chekhov's footgun with `#[non_exhaustive]` enums.


> I find conditional enum variants problematic:
> 
> * One might accidentally not handle that variant when developing on a system where that variant is disabled. The error then only gets caught downstream (or in CI) not at compile time **on the developers machine**.

Could you please elaborate this a bit more in detail? Do you have examples where this gets particularly annoying? May be I'm too much into cross-platfom work with serialport to immediately see big shortcommings because my default is to cross-compile locally a lot.


> * Every match arm matching a conditional variant now needs to repeat the conditional.
> 
> One alternative would be to make those variants not constructible but still nameable when unsupported. E.g. by making it a `#[non_exhaustive]` struct/tuple and adding a field of a valueless type when unsupported. Though this requires adding a constructor function for supported targets.

This is from my perspective one of the two big down-sides. And the other being forgetting to match a variant due to `#[non_exhaustive]` always requiring a default match arm. But only a few of the direct dependencies perform matches at all and I would prefer placing the burden on the ones matching these types instead of the ones which just set up parity or stop bits. 


> Playground example for `StopBits` enum:
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d33554945a7d86532d6a20feeeba21c6

Thank you for this illustrative example! My gut instinct rebels against the different syntax variants for constructing enum instances. But luckily as `Parity::OnePointFive` is an empty struct, it can be instantiated just as `Parity::OnePointFive` as in [this derived example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=098d624917a49b3b407ba8aa80b10390). Shouldn't this do the trick for our use case? Treating `Parity::OnePointFive{..}` would be only required when matching which is fine for me with respect to where to put the additonal burden.

Skgland:
> Thank you for preparing this PR @NONnonHere! And thank you for your feedback and pointing out an interresting potential alternative @Skgland! I see your point and Chekhov's footgun with `#[non_exhaustive]` enums.
> 
> > I find conditional enum variants problematic:
> > 
> > * One might accidentally not handle that variant when developing on a system where that variant is disabled. The error then only gets caught downstream (or in CI) not at compile time **on the developers machine**.
> 
> Could you please elaborate this a bit more in detail? Do you have examples where this gets particularly annoying? May be I'm too much into cross-platfom work with serialport to immediately see big shortcommings because my default is to cross-compile locally a lot.

I ran into this problem in an internal project with an enum that had variants configured based on selected features and targets.
Some dependencies were troublesome to cross-compile so that wasn't yet an option.
Also I generally try to have as few code as possible removed via `#[cfg(...)]` and usually try to use `cfg!(...)` so that everything at least gets syntax and type checked independent of the current host/target.

Given the size of the enum, the count of cfg-ed out variants and there rather simple cfg-conditions I might be making a mountain out of a mole-hill here.

> 
> > * Every match arm matching a conditional variant now needs to repeat the conditional.
> > 
> > One alternative would be to make those variants not constructible but still nameable when unsupported. E.g. by making it a `#[non_exhaustive]` struct/tuple and adding a field of a valueless type when unsupported. Though this requires adding a constructor function for supported targets.
> 
> This is from my perspective one of the two big down-sides. And the other being forgetting to match a variant due to `#[non_exhaustive]` always requiring a default match arm. But only a few of the direct dependencies perform matches at all and I would prefer placing the burden on the ones matching these types instead of the ones which just set up parity or stop bits.

Yeah, having to call a cfg-ed constructor function rather than just using the enum variant is a little bit annoying.
Though they still need to either add the cfg in their code either way or only support targets were the variant is available. 

> 
> > Playground example for `StopBits` enum:
> > https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d33554945a7d86532d6a20feeeba21c6
> 
> Thank you for this illustrative example! My gut instinct rebels against the different syntax variants for constructing enum instances. But luckily as `Parity::OnePointFive` is an empty struct, it can be instantiated just as `Parity::OnePointFive` as in [this derived example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=098d624917a49b3b407ba8aa80b10390). Shouldn't this do the trick for our use case? Treating `Parity::OnePointFive{..}` would be only required when matching which is fine for me with respect to where to put the additonal burden.

I think this only works because the enum is defined in the same crate that it is used in.
Cross-crate I would expect the `#[non_exhaustive]` to prevent the variant from being constructed.
Honestly I am surprised that this works at all constructing a struct variant like a unit variant.

PR #366, opened 2026-07-30
feat: add a 9-bit sized character option

PureTryOut:
Used by some embedded applications.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -153,6 +153,9 @@ pub enum DataBits {
 
     /// 8 bits per character
     Eight,
+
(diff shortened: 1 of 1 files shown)


sirhcel:
Thank you for starting this PR @PureTryOut! I see the use-case. But I'm wondering whether this would make the current API more brittle because:

1. This PR only adds `DataBits::Nine` but does not handle it anywhere
2. We have no transport for data wider than eight bits and would send a not obviously defined ninth bit and silently drop in for received data

Where are you heading to with this PR in it's current state?

PR #367, opened 2026-08-17
return errors instead of panicking when reading baud rate

NONnonHere:
`baud_rate()` can terminate the operation in a function that is returning `Result`:

- `assert!(ospeed == ispeed)` in three platform variants — input/output speed
  can be different on the communication line, and the device itself can report it
- `_ => unreachable!()` closing the speed matching for PowerPC Linux, reachable for
  any port set to a rate outside the hardcoded list

Both now return `ErrorKind::Unknown`. Adds `## Errors` doc sections to match.


src/posix/tty.rs
```diff
@@ -517,7 +522,12 @@ impl TTYPort {
...
 
-        assert!(termios2.c_ospeed == termios2.c_ispeed);
+        if termios2.c_ospeed != termios2.c_ispeed {
+            return Err(Error::new(
+                ErrorKind::Unknown,
+                "port reports differing input and output baud rates",
+            ));
+        }
 
         Ok(termios2.c_ospeed)
```
sirhcel: What about going one step further and cleaning up this decision to the point where we no longer need a return? For example:

```suggestion
        if termios2.c_ospeed == termios2.c_ispeed {
            Ok(termios2.c_ospeed)
        } else {
            Err(Error::new(
                ErrorKind::Unknown,
                "port reports differing input and output baud rates",
            ))
        }
```

src/posix/tty.rs
```diff
@@ -537,7 +552,12 @@ impl TTYPort {
...
 
-        assert!(ospeed == ispeed);
+        if ospeed != ispeed {
+            return Err(Error::new(
+                ErrorKind::Unknown,
+                "port reports differing input and output baud rates",
+            ));
+        }
 
         Ok(ospeed as u32)
```
sirhcel: Same here:

```suggestion
        if ospeed == ispeed {
            Ok(ospeed as u32)
        } else {
            Err(Error::new(
                ErrorKind::Unknown,
                "port reports differing input and output baud rates",
            ))
        }
```

src/posix/tty.rs
```diff
@@ -606,7 +637,15 @@ impl TTYPort {
...
-            _ => unreachable!(),
+            _ => {
+                return Err(Error::new(
+                    ErrorKind::Unknown,
+                    format!(
+                        "port reports an unrecognized baud rate constant: {}",
+                        ospeed
+                    ),
+                ))
+            }
```
sirhcel: What about factoring this entire match statement out into an individual function and letting the function originally containing it end with a single if/else statment like in the cases above?

sirhcel (commented):
Thank you for cleaning this up! I would go one step further and convert the first two places to a single if/else statement. From my perspective this makes the intent more clear because the comparison is no longer negated and the decision becomes a single if/else statement.

The third place is too big for doing the same in-place for my taste. But may be factoring the match out into a function will give us the same brevity.

sirhcel:
Superseded by #368 against main.

Issue #348, opened 2026-05-06
usb: provide port chain and bus ID of port


xobs:
While we hope  that most devices provide a unique serial number, many products just leave this blank. This is particularly true with serial ports that are embedded in unrelated devices to control them.

As an example, I have two hubs with a small serial port that you can connect to and control power to the ports.

If `serialport` provided the [port chain](https://docs.rs/nusb/latest/nusb/struct.DeviceInfo.html#method.port_chain) and [bus id](https://docs.rs/nusb/latest/nusb/struct.DeviceInfo.html#method.bus_id) we could use that to disambiguate different devices, as well as associate a port with a particular bus topology.

The `nusb` crate does this in a cross-platform manner, so perhaps it could be used as an example.

sirhcel:
This is a definitely a good point @xobs! We had some discussion on that topic in the past in #222 and #227 in the past.

I have not looked in the details of port chains on different platforms. But I'm assuming that the nusb crate works across different platforms and we would use their representation of port ID as `&str` and the port chain as `&[u8]`.

xobs:
I thought I did a search, but I guess I missed that issue. Thanks for the pointer to it.

PR #369, opened 2026-08-24
Release 4.10

sirhcel:
We've added two new features:

* USB port chains information (#348)
* Target support for illumos (#342 

And on top we are returning errors when reading back inconsistent baud rates from the OS on some POSIX targets (#367). This looks like a good reason for cutting a new release.

Semver checks currently fail due to the added USB port chains. Even then when they gated are behind a preview feature. This will get back to normal after tagging the release.


eldruin (approved):
Sounds good to me, thank you!

sirhcel:
Thank you very much for having a look @eldruin!

I rebased for including the fix from #370 as last minute addition. Build passes and the new test for parsing serial numbers with dashes passes in CI.

Issue #351, opened 2026-05-28
Trimmed Serial Number in Windows


cruzjuniel:
I am working on a custom USB Serial Device that uses a custom formatted Serial Number. The serial number is correct when using macOS. But not on Windows.

If it helps, the serial number starts with a prefix followed by a hyphen/dash (-) character. For example: `ABC-DE-FGHIJ012345` which returns completely on macOS but on Windows, it is only read as `ABC`. It seems that the string is getting trimmed at the first (-) character.

MRWGNRRR:
+1

<img width="864" height="600" alt="Image" src="https://github.com/user-attachments/assets/8e8fdd56-381c-4510-902c-fe01f07fb3d5" />

<img width="511" height="400" alt="Image" src="https://github.com/user-attachments/assets/f33de2c9-6ff1-4ef3-8dfe-183d21e01234" />

sirhcel:
Thank you for reporting this issue @cruzjuniel and @MRWGNRRR! We plainly did not accept the dash as part of the serial number, #370 fixes this specific case, and I snug this fix into today's release [4.10.0](https://github.com/serialport/serialport-rs/releases/tag/v4.10.0).

At a first glance, there is more to consider for getting serial number parsing right. More characters should be accepted and we should likely refactor for parsing different variants like USB in general and FTDI (#178, where an interface identifier/index is appended to the serial number). So I will leave this issue open for now.

MRWGNRRR:
> Thank you for reporting this issue [@cruzjuniel](https://github.com/cruzjuniel) and [@MRWGNRRR](https://github.com/MRWGNRRR)! We plainly did not accept the dash as part of the serial number, [#370](https://github.com/serialport/serialport-rs/pull/370) fixes this specific case, and I snug this fix into today's release [4.10.0](https://github.com/serialport/serialport-rs/releases/tag/v4.10.0).
> 
> At a first glance, there is more to consider for getting serial number parsing right. More characters should be accepted and we should likely refactor for parsing different variants like USB in general and FTDI ([#178](https://github.com/serialport/serialport-rs/issues/178), where an interface identifier/index is appended to the serial number). So I will leave this issue open for now.

what about reading device name?
at the moment, the crate only reads FriendlyName

PR #370, opened 2026-08-26
Accept dash in serial numbers on Windows

sirhcel:
This fixes the specific case from https://github.com/serialport/serialport-rs/issues/351. But it looks like there is more to consider to get serial number parsing right.


sirhcel:
Semver tests are failing due to unreleased feature gate from #350.

PR #372, opened 2026-08-27
haiku: Add support for Haiku

kallisti5:


kallisti5:
The nix crate is missing a few needed ioctl_ wrappers on Haiku, so we have a few minor hacks to make things work for the short term until we can get fixes in upstream nix.

